PR #1, opened 2022-01-19
Add macOS support using Core Graphics

sanxiyn:
It draws something, but I am not confident it is correct. I am especially worried about coordinate system.

What should I expect to see for winit example? I think it would help to include screenshots in README (plural, to show resizing), and for example to be recognizably different when horizontally or vertically flipped. I propose a 45-60-75 triangle.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,10 @@ x11-dl = "2.19.1"
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = "0.3.9"
 
+[target.'cfg(target_os = "macos")'.dependencies]
+core-graphics = "0.22.3"
+objc = "0.2.7"
+
 [dev-dependencies]
 winit = "0.26.1"
 image = "0.23.14"
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -1,0 +1,55 @@
+use crate::{GraphicsContextImpl, SoftBufferError};
+use raw_window_handle::{HasRawWindowHandle, AppKitHandle};
+use objc::runtime::Object;
+use core_graphics::base::{kCGBitmapByteOrder32Little, kCGImageAlphaNoneSkipFirst, kCGRenderingIntentDefault};
+use core_graphics::color_space::CGColorSpace;
+use core_graphics::context::CGContext;
+use core_graphics::data_provider::CGDataProvider;
+use core_graphics::geometry::{CGPoint, CGSize, CGRect};
+use core_graphics::image::CGImage;
+use core_graphics::sys;
+
+pub struct CGImpl;
+
+impl CGImpl {
+    pub unsafe fn new<W: HasRawWindowHandle>(handle: AppKitHandle) -> Result<Self, SoftBufferError<W>> {
+        let window = handle.ns_window as *mut Object;
+        let cls = class!(NSGraphicsContext);
+        let graphics_context: *mut Object = msg_send![cls, graphicsContextWithWindow:window];
+        if graphics_context.is_null() {
+            return Err(SoftBufferError::PlatformError(Some("Graphics context is null".into()), None));
+        }
+        let _: () = msg_send![cls, setCurrentContext:graphics_context];
+        Ok(Self)
+    }
+}
+
+impl GraphicsContextImpl for CGImpl {
+    unsafe fn set_buffer(&mut self, buffer: &[u32], width: u16, height: u16) {
+        let cls = class!(NSGraphicsContext);
+        let graphics_context: *mut Object = msg_send![cls, currentContext];
+        let context_ptr: *mut sys::CGContext = msg_send![graphics_context, CGContext];
+        let context = CGContext::from_existing_context_ptr(context_ptr);
+        let color_space = CGColorSpace::create_device_rgb();
+        let slice = std::slice::from_raw_parts(
+            buffer.as_ptr() as *const u8,
+            buffer.len() * 4);
+        let data_provider = CGDataProvider::from_slice(slice);
+        let image = CGImage::new(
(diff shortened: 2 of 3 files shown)


john01dav:
I've added a new fruit example to show a proper image that I got off Wikimedia Commons. That should test a bit more thoroughly than a simple triangle since it has lots of colors and is otherwise more complicated so more can go wrongly. I'll probably add something to the README later, but I wanted to get this to you quickly.

Here are some screenshots (2 on Fedora Linux and two on Windows 10, note that when the window is larger than the displayed image it doesn't matter what is displayed and is different between Linux and Windows):


Screenshot from 2022-01-18 22-05-25
Screenshot from 2022-01-18 22-05-53
Screenshot from 2022-01-18 22-06-26
Screenshot from 2022-01-18 22-06-38

john01dav:
Here is the winit example on both Linux and Windows. Note which side of the squares each color shows up on:
Screenshot from 2022-01-18 22-12-21
Screenshot from 2022-01-18 22-14-24

sanxiyn:
Thanks! That was helpful. I am now confident this works correctly.

john01dav:
Thanks for the PR!

john01dav:
@sanxiyn I'm updating the README to reflect the new AppKit support. Do you have a preference for what name you are credited with?

sanxiyn:
Please use my Git user.name, that is Seo Sanghyeon.

john01dav:
@sanxiyn I recently got access to a Mac computer of my own (albeit an old one, it supports nothing after High Sierra) and I was testing some related code on it and I found some cases in your implementation that don't work correctly. I was able to solve one of them (drawing an empty image causes an assertion to fail, so it's easy enough to fix by simply not drawing the image if width and height are zero), but I was unable to fix the other.

It seems that once an image is drawn to the window, it is impossible to draw to it again with a new image until after the window resizes. I tried to fix this, but my unfamiliarity with the core graphics API made this not work out. I'll keep trying, but you might have a better chance to figure it out.

I've added a new animation example that reliably reproduces the bug on Mac OS. [Here]([animation-example.zip](https://github.com/john01dav/softbuffer/files/7975797/animation-example.zip))'s a video of what it is supposed to look like. I had to put it in a zip so Github would accept the upload. This example also lags pretty hard on Mac OS (but not on Windows or Linux, except when resizing as it blocks the main thread to re-render the animation, but even if that is an issue it's a winit one or an issue with the example and not a softbuffer one -- it lags on Mac OS when `set_buffer` alone is spammed).

kpreid:
> It seems that once an image is drawn to the window, it is impossible to draw to it again with a new image until after the window resizes.

I'm seeing the same behavior you describe on macOS Monterey 12.2 (with `cargo run --example animation --release`, or with my own code).

Unfortunately I also don't have any ideas about how to fix it, or at least not ones that worked. I was wondering if it was missing some sort of “flush” operation, but `msg_send![graphics_context, flushGraphics]` and `context.flush();` changed nothing. Accessing/creating the graphics context anew for each `set_buffer` didn't work entirely. Another idea I had was that maybe the graphics context wasn't available as intended (perhaps set to clip everything) except when resizing, but if that's so, I don't see what to change to fix that, short of implementing a “proper” NSView subclass. (I used to work with AppKit years ago, but never Core Graphics.)

john01dav:
Something else that might be relevant is this (
https://github.com/rust-windowing/winit/issues/1605), but unfortunately my
attempt to use it didn't work (described in the linked issue). Maybe this
means more to you? If that doesn't work I'll probably have to contact Apple
(and who knows if they'll be helpful), or give up and have a GPU dependent
buffer on Mac OS, as minifb does.

kpreid:
I'm not quite following. You say it doesn't work here, but when I follow the link I see something that says it _would_ work if winit was changed to call `setNeedsDisplay`? (Which sounds like a fine idea to me — except that it might be an extra cost to applications that don't need it, and it means that softbuffer would need to document that drawing _must_ be done from RedrawRequested, rather than that it _can_ be done there.)

Another likely correctness issue here is that the current `set_buffer()` code uses the thread current graphics context rather than one obtained from the window and remembered or re-fetched; this would surely draw in the wrong place if there are multiple windows, and I suspect the current behavior has to do with the state of that context. However, I tried to fix that and both of the approaches I tried resulted in no drawing at all rather than the current sometimes-drawing.

Maybe the GPU dependent drawing is the better answer all around, since it means you get a layer of your own independent of what the AppKit window drawing is doing.

john01dav:
It turned out that I had missed something obvious in my testing of `setNeedsDisplay`. It was actually working, but because I had slowed the update rate to 1Hz and because the animation is one second long it was showing the same frame each time and thus appearing to not work. Once I increased the animation speed back to where it should be, everything works. I have now made [a PR to winit](https://github.com/rust-windowing/winit/pull/2189) about `setNeedsDisplay`.

tinaun:
I do want to try an alternative pr using a custom layer backed NSView at some point

john01dav:
@tinaun I always want the code to be as correct as possible, but I am unsure what the advantage is of doing it that way. Is it just a general sense of correctness (and there's nothing wrong with this), or is there some more practical advantage? Please don't take this to be discouraging in any way. I'm just trying to understand your motivation and any potential shortcomings of the current solution.

tinaun:
1) this will (hopefully) support scaling in retina modes better, as the current examples always draw the images in logical pixels not physical pixels

2) your pr for using setNeedsDisplay in winit breaks opengl/metal

3) that was the approach I was going to try before knowing this repo existed :v

john01dav:
@tinaun Those are great reasons, thank you for clarifying. If you make the PR I'll definitely merge it.

PR #2, opened 2022-02-12
Add web support

Liamolucko:
Adds web support via. `CanvasRenderingContext2D.putImageData`,

A limitation of this implementation is that it can't coexist with anything which creates a canvas context other than `CanvasRenderingContext2D`, since only one context can exist per canvas.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,13 @@ winapi = "0.3.9"
 core-graphics = "0.22.3"
 objc = "0.2.7"
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = "0.2.78"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies.web-sys]
+version = "0.3.55"
+features = ["CanvasRenderingContext2d", "Document", "Element", "HtmlCanvasElement", "ImageData", "Window"]
+
 [dev-dependencies]
 winit = "0.26.1"
 image = "0.23.14"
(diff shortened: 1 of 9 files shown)


Liamolucko:
The examples don't need to be made into libraries to be used; `wasm-bindgen` shims will automatically run `main`. (You might be thinking of `wasm-pack`, which can only build libraries.)
To use the compiled WASM, you need to run it through `wasm-bindgen` to create a JS shim, and then use a HTML file which imports it, e.g.:
```rust
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
</head>
<body>
    <script type="module">
        import init from "./<wasm-bindgen out dir>/<example name>.js";
        init();
    </script>
</body>
</html>
```

[`cargo-run-wasm`](https://github.com/rukai/cargo-run-wasm) can do all that automatically, so that might be a good choice.

john01dav:
Thanks for clarifying. I was able to get some things running that way, but there is still an issue. The animation example depends on Rayon (I modified it to not use Rayon to get it working), as does fruit indirectly through the `image` crate. Can you clarify how you tested these without such changes, or, if you did make such changes, why they are not committed (even as a separate example, e.g. fruit_wasm and animation_wasm)?

Given how much trouble I'm having making the examples work, it might be prudent to add a section in README.md explaining how to do so. Examples are there to help people learn, so providing such information seems reasonable. What do you think?

john01dav:
With your changes to remove rayon I can now get the examples to work, so I'm going to go ahead and merge this. I'd like to see a section added to README.md documenting how to run the examples, but it's useful to have this merged before that is available. Feel free to submit another PR if you'd like to write that documentation, otherwise I will at some point.

PR #3, opened 2022-04-24
use CALayer for macOS backend

lunixbochs:
This should fix: https://github.com/rust-windowing/winit/pull/2189 and https://github.com/rust-windowing/winit/issues/1605

As you're already constructing a `CGImage`, you can create an `NSImage` backed by it using `NSCGImageRep`

You can set your `view.layer` to a new `CALayer`, and repeatedly set `layer.contents` to the new `NSImage` instead of trying to draw to the `NSView` directly.

This approach works whether or not you're called from a draw callback, and does not require any changes to `winit`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,9 @@ x11-dl = "2.19.1"
 winapi = "0.3.9"
 
 [target.'cfg(target_os = "macos")'.dependencies]
+cocoa = "0.24.0"
 core-graphics = "0.22.3"
+foreign-types = "0.3.0"
 objc = "0.2.7"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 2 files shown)


john01dav:
Sorry for taking so long to review this. I've been very busy lately, and my Mac OS computer wasn't deployed so it took some time to do the review. I have some concerns. When I ran this with the animation example/test, it worked perfectly, but there are some concerns on the other examples/tests. The two issues that I noticed at that 1) resizing the window no longer fires a repaint event, and 2) when the size of the buffer to display does not match the window's size the buffer is stretched rather than being placed in the upper-left corner as on other platforms as the documentation specifies. We can change the documentation to permit this different behavior, but I would prefer to not have such behavior vary between platforms as that's likely to reduce bugs in practice. Is this feasible with your new way of doing this?

Also, thank you so much for this fix. It's been hard to get help from anyone who really knows Mac OS's APIs well, so I *very* much welcome this improvement. I'll merge it as-is with the documentation changes if needed. Please don't take how long it took me to review to mean that I do not care about this PR ­— it was just a case of getting my Mac OS computer re-deployed in order to properly review it.

Liamolucko:
> 1) resizing the window no longer fires a repaint event

This is probably the same issue as gfx-rs/wgpu#2144, which is an upstream bug in winit; there's a bit more discussion in gfx-rs/wgpu#2543.

john01dav:
This works great now, thanks for the PR! It is now merged. I'll push a new version on crates.io soon. How would you like me to credit you in the README.md alongside the author of the original Mac OS implementation?

PR #5, opened 2022-05-26
Make `GraphicsContext::window_mut()` an `unsafe fn`.

kpreid:
As documented in the new safety comment, providing `&mut` access to an inner component about which there are consistency invariants is unsafe, because `&mut` is sufficient for a caller to completely replace the value (using assignment or `std::mem::swap()`).

Luckily, when `softbuffer` is used with `winit`, no `&mut` access is needed. However, other windowing libraries such as `sdl2` do have `&mut` methods, so this method can't simply be removed. (I don't know which windowing libraries would in fact be compatible with softbuffer and thus relevant.)

This is a breaking change since it makes a previously safe function unsafe, and should not be published without a major version bump (i.e. to `0.2.0` or higher).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,15 +62,28 @@ impl<W: HasRawWindowHandle> GraphicsContext<W> {
         })
     }
 
-    /// Gets shared access to the underlying window
+    /// Gets shared access to the underlying window.
     #[inline]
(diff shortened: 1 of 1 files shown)


john01dav (approved):

john01dav:

PR #6, opened 2022-05-26
Add `impl AsRef<W> for GraphicsContext<W>`.

kpreid:
This will allow code which works with windows generically (such as an event loop which can work with `softbuffer` or another graphics library) to be able to access the underlying window without knowing about `softbuffer` in particular.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -111,6 +111,7 @@ impl<W: HasRawWindowHandle> GraphicsContext<W> {
 
 impl<W: HasRawWindowHandle> AsRef<W> for GraphicsContext<W> {
     /// Equivalent to [`self.window()`](Self::window()).
+    #[inline]
     fn as_ref(&self) -> &W {
         self.window()
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -109,6 +109,13 @@ impl<W: HasRawWindowHandle> GraphicsContext<W> {
     }
 }
 
+impl<W: HasRawWindowHandle> AsRef<W> for GraphicsContext<W> {
+    /// Equivalent to [`self.window()`](Self::window()).
+    fn as_ref(&self) -> &W {
```
john01dav, line 114: Is there any particular reason that this isn't `#[inline]`?

john01dav:

src/lib.rs
kpreid: No; done.

This may be redundant since the function is generic and generic functions are, in the current Rust compiler, always inlinable, but it makes some sense to _explicitly_ follow the general recommended practice of making trivial accessor functions inlinable.

john01dav:
Thanks for the change!

PR #7, opened 2022-05-26
Apply rustfmt and other formatting tweaks.

kpreid:
I don't know if this change is welcome, but when I noticed format-on-save in my editor changing some existing formatting, I figured it's worth checking.

There are two commits in this PR:

* The first one merely applies `cargo fmt`. I reviewed the results to verify that readability was not harmed; the only thing I noticed that might be undesirable was `std` imports getting sorted alphabetically in their group. This could be fixed by inserting a blank line.

* The second one rewraps comments and string literals so that they fit in the standard 100-column line length. (`rustfmt` skips formatting lines that contain such literals and comments, so this is helpful for ensuring standard formatting of the code surrounding them.)

### DIFF
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -1,8 +1,8 @@
-use std::f64::consts::PI;
 use instant::Instant;
 #[cfg(not(target_arch = "wasm32"))]
 use rayon::prelude::*;
 use softbuffer::GraphicsContext;
+use std::f64::consts::PI;
 use winit::event::{Event, WindowEvent};
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::WindowBuilder;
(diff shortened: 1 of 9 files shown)


john01dav:
This is a great change, and very welcome, thank you! Normally I have my environment configured to run `cargo fmt` before every build/run/check/test run, but there's a lot of platform hopping to work on this, so I didn't fully configure all applicable environments. I'm not going to merge this now though, since there are other PRs that this may unnecessarily create merge conflicts with. Instead, I'll run `cargo fmt` myself, with `wrap_comments = true` in rustfmt.toml. If I forget, feel free to ping me on Discord, open an issue, or some other such thing and I'll do it then. his option requires nightly to reformat, but that's worth it to automate the comment wrapping I think. I would be much more concerned about it if nightly were required to build the formatted code.

Issue #9, opened 2022-05-31
No way to damage surface in Wayland


i509VCB:
Many compositors implement damage tracking, meaning that commiting a new buffer when calling set_buffer does nothing.

A proper fix would involve the caller giving the implementation a set of rectangles where content has changed.

john01dav:
Closing as duplicate of #39 with the merge of swbuf and softbuffer

PR #12, opened 2022-05-31
wayland: temporary horrible fix for damage

i509VCB:
This is NOT a permanent fix to #9.

Fixing #9 will require some additional api changes.

At the moment we do not damage the surface on wayland. This means calling set_buffer will do nothing on some compositors (KDE in my case).

Damage is also broken with window scaling, the API will need to be adjusted to add damage.


john01dav:
Thanks for the PR! Sorry for the delay. I had some employment contract issues with FOSS contributions, and they are now resolved.

PR #21, opened 2022-07-01
Enable needed features of winapi

Liamolucko:
Previously, `softbuffer` wouldn't successfully build on its own on Windows; the examples only worked because dependencies happened to enable the needed features.


john01dav:
Thanks for the patch! Sorry for the delay. I had some employment contract issues with FOSS contributions, and they are now resolved.

PR #25, opened 2022-08-12
Update raw_window_handle to 0.5 and winit to 0.27.1

lucasmerlin:
To work with winit 0.27, raw_window_handle needs to be updated to 0.5. I've tested the examples, they all still run successfully on my mac.


john01dav:
I tried to build this for Windows and Linux, and with your change it does not build. I'll likely end up just doing this update myself, as it is very important. 

john01dav:
I have now pushed my own version of this change.

PR #32, opened 2022-12-13
   Fix scaling of the buffer on macOS

### LINKED ISSUE #22: Support for physical vs logical pixels
On macOS, the buffer is applied using logical pixels rather than physical pixels so a 100x100 buffer will fill a 100x100 (logical pixels) window even if the window occupies 200x200 (physical) pixels. A 200x200 buffer in the same scenario will just be cropped. This is a good default that makes UI easier to develop, but results in output that looks low res.

I'm currently working around this by setting my buffer to the physical pixel size, then scaling it on application by adding the following to set_buffer in cg.rs:
`self.layer.set_contents_scale(width as f64 / self.layer.bounds().size.width);`

I'm not sure what the right and real mechanism for doing this should be (e.g. always use physical? allow scale factor choice?)

tronical:
Currently the size of the buffer on macOS is interpreted in logical pixels, which
is inconsistent with the other platforms. Instead,
we should apply the same scale factor that applies to regular rendering.

Fixes #22

### DIFF
--- a/examples/fruit.rs
+++ b/examples/fruit.rs
@@ -16,7 +16,10 @@ fn main() {
     }).collect::<Vec<_>>();
 
     let event_loop = EventLoop::new();
-    let window = WindowBuilder::new().build(&event_loop).unwrap();
+    let window = WindowBuilder::new()
+        .with_inner_size(winit::dpi::PhysicalSize::new(fruit.width(), fruit.height()))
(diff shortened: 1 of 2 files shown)


tronical:
Moving this to https://github.com/rust-windowing/swbuf/pull/6

PR #35, opened 2022-12-20
Improvements to wayland backend, and build on FreeBSD

ids1024:
- Use wayland-client `0.30.0-beta.14`
  * The stable release for 0.30.0 is hopefully not too far away now: https://github.com/Smithay/wayland-rs/issues/431
- Use `memfd` instead of temporary file
- Make sure buffer and buffer pool objects are destroyed
- Build on FreeBSD
  * Apparently other BSDs don't support memfd? Not sure what the best solution is there.

### DIFF
--- a/src/wayland.rs
+++ b/src/wayland.rs
@@ -1,74 +1,140 @@
+use crate::{error::unwrap, GraphicsContextImpl, SoftBufferError};
+use nix::sys::memfd::{memfd_create, MemFdCreateFlag};
 use raw_window_handle::{HasRawWindowHandle, WaylandDisplayHandle, WaylandWindowHandle};
-use tempfile::tempfile;
-use wayland_client::{Display, sys::client::wl_display, GlobalManager, protocol::{wl_shm::WlShm, wl_buffer::WlBuffer, wl_surface::WlSurface}, Main, Proxy, EventQueue};
-use crate::{GraphicsContextImpl, SoftBufferError, error::unwrap};
-use std::{fs::File, os::unix::prelude::{AsRawFd, FileExt}, io::Write};
+use std::{
+    ffi::CStr,
+    fs::File,
+    io::Write,
+    os::unix::prelude::{AsRawFd, FileExt, FromRawFd},
+};
+use wayland_client::{
+    backend::{Backend, ObjectId},
+    globals::{registry_queue_init, GlobalListContents},
+    protocol::{wl_buffer, wl_registry, wl_shm, wl_shm_pool, wl_surface},
+    Connection, Dispatch, EventQueue, Proxy, QueueHandle,
+};
+
+struct State;
 
 pub struct WaylandImpl {
-    _event_queue: EventQueue,
-    surface: WlSurface,
-    shm: Main<WlShm>,
+    event_queue: EventQueue<State>,
+    qh: QueueHandle<State>,
+    surface: wl_surface::WlSurface,
+    shm: wl_shm::WlShm,
     tempfile: File,
-    buffer: Option<WaylandBuffer>
+    buffer: Option<WaylandBuffer>,
 }
 
-struct WaylandBuffer{
+struct WaylandBuffer {
     width: i32,
     height: i32,
-    buffer: Main<WlBuffer>
+    pool: wl_shm_pool::WlShmPool,
+    buffer: wl_buffer::WlBuffer,
+}
+
+impl Drop for WaylandBuffer {
+    fn drop(&mut self) {
+        self.buffer.destroy();
+        self.pool.destroy();
+    }
 }
 
 impl WaylandImpl {
-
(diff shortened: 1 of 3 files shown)


src/wayland.rs
```diff
@@ -1,84 +1,201 @@
...
+            registry_queue_init(&conn),
+            "Failed to make round trip to server",
+        )?;
+        let qh = event_queue.handle();
+        let shm: wl_shm::WlShm = unwrap(
+            globals.bind(&qh, 1..=1, ()),
+            "Failed to instantiate Wayland Shm",
+        )?;
+        let name = CStr::from_bytes_with_nul_unchecked("softbuffer\0".as_bytes());
+        let tempfile_fd = unwrap(
```
i509VCB, line 60: We should have a fallback or else this only works on Freebsd and Linux. See sctk for the fallback.

src/wayland.rs
```diff
@@ -1,84 +1,201 @@
...
-        unwrap(event_queue.sync_roundtrip(&mut (), |_, _, _| unreachable!()), "Failed to make round trip to server")?;
-        let shm = unwrap(globals.instantiate_exact::<WlShm>(1), "Failed to instantiate Wayland Shm")?;
-        let tempfile = unwrap(tempfile(), "Failed to create temporary file to store buffer.")?;
-        let surface = Proxy::from_c_ptr(window_handle.surface as _).into();
-        Ok(Self{
-            _event_queue: event_queue,
-            surface, shm, tempfile,
-            buffer: None
+impl WaylandImpl {
+    pub unsafe fn new<W: HasRawWindowHandle>(
```
i509VCB, line 43: I think we should deny `unsafe_op_in_unsafe_fn` in the whole crate.

src/wayland.rs
```diff
@@ -1,84 +1,201 @@
...
+
+impl Dispatch<wl_buffer::WlBuffer, ()> for State {
+    fn event(
+        _: &mut State,
+        _: &wl_buffer::WlBuffer,
+        _: wl_buffer::Event,
+        _: &(),
+        _: &Connection,
+        _: &QueueHandle<State>,
+    ) {
```
i509VCB, line 199: We need to handle release in the future, this needs a todo comment.

src/wayland.rs
```diff
@@ -1,84 +1,201 @@
...
+        let (globals, event_queue) = unwrap(
+            registry_queue_init(&conn),
+            "Failed to make round trip to server",
+        )?;
+        let qh = event_queue.handle();
+        let shm: wl_shm::WlShm = unwrap(
+            globals.bind(&qh, 1..=1, ()),
+            "Failed to instantiate Wayland Shm",
+        )?;
+        let name = CStr::from_bytes_with_nul_unchecked("softbuffer\0".as_bytes());
```
i509VCB, line 59: Couldn't this fail if a file of the same name exists?

src/wayland.rs
```diff
@@ -1,84 +1,201 @@
...
+        ));
+        let (globals, event_queue) = unwrap(
+            registry_queue_init(&conn),
+            "Failed to make round trip to server",
+        )?;
+        let qh = event_queue.handle();
+        let shm: wl_shm::WlShm = unwrap(
+            globals.bind(&qh, 1..=1, ()),
+            "Failed to instantiate Wayland Shm",
+        )?;
```
i509VCB, line 58: We should do a roundtrip after this to ensure it is initialized.

src/wayland.rs
```diff
@@ -1,84 +1,201 @@
...
     }
 
-    fn ensure_buffer_size(&mut self, width: i32, height: i32){
-        if !self.check_buffer_size_equals(width, height){
-            let pool = self.shm.create_pool(self.tempfile.as_raw_fd(), width*height*4);
-            let buffer = pool.create_buffer(0, width, height, width*4, wayland_client::protocol::wl_shm::Format::Xrgb8888);
-            self.buffer = Some(WaylandBuffer{
+    fn ensure_buffer_size(&mut self, width: i32, height: i32) {
+        if !self.check_buffer_size_equals(width, height) {
+            let pool =
```
i509VCB, line 88: This will cause an fd leak, we don't destroy the old pool of it did exist.

src/wayland.rs
```diff
@@ -1,84 +1,201 @@
...
+            "Failed to make round trip to server",
+        )?;
+        let qh = event_queue.handle();
+        let shm: wl_shm::WlShm = unwrap(
+            globals.bind(&qh, 1..=1, ()),
+            "Failed to instantiate Wayland Shm",
+        )?;
+        let name = CStr::from_bytes_with_nul_unchecked("softbuffer\0".as_bytes());
+        let tempfile_fd = unwrap(
+            memfd_create(name, MemFdCreateFlag::MFD_CLOEXEC),
```
i509VCB, line 61: Should we seal the file if supported by the platform to be nice to compositors?

ids1024:
This was merged into https://github.com/rust-windowing/swbuf, which is now incorporated into this repository.

Issue #38, opened 2022-12-21
Use a shared memory buffer for the X11 backend


notgull:
The current X11 backend sends the image over the wire, which is slow and inefficient. We should probably use the [MIT-SHM](https://www.x.org/releases/X11R7.7/doc/xextproto/shm.html) strategy for this crate if it's available.

(Also, we probably shouldn't create a new image every time we run the `set_buffer` command, but I think we can take care of that when we solve rust-windowing/softbuffer#37)

Issue #40, opened 2022-12-21
Make softbuffer window contexts own the pixel buffers.


i509VCB:
At the moment all backends need to perform a copy to present. Some backends such as Wayland (always) and X11 (with the SHM extensions) can simply share the pixel memory with the server. On Windows and Mac this saving wouldn't be possible, but it would help on other platforms.

Not sure what redox does in this case.

notgull:
Given that we're probably moving in this direction, I figure we should come up with an API for this. Here's some considerations:

* How do we expose the buffer to the user? Do we have a `buffer()/buffer_mut()` method set that directly returns a slice? Or, do we want to make access more restrictive (maybe a `with_buffer` method that takes a closure that operations on a slice).
* How do we handle resizing the buffer?
* How should the buffer be pushed to the server?

ids1024:
For the Wayland backend, at least, we could possibly hand out an owned object that `DerefMut`s to `[u8]`. Then the client transfers it back to display the buffer on the screen. That's basically how the current `WaylandBuffer` type works. But maybe that doesn't make as much sense in other contexts, and isn't actually a more convenient API.

I think `buffer`/`buffer_mut` is fine and `with_buffer` is unnecessary. As long as the method that submits/swaps the buffer takes `&mut self`, that will ensure the caller no longer has any references to the buffer.

ids1024:
Another consideration is that `set_buffer` takes a width/height, while that's needed to allocate/resize buffers before we can provide a slice.

One possible API for this, and https://github.com/rust-windowing/softbuffer/issues/37:
```rust
Context::new<T: HasRawDisplayHandle>(display: &T) -> Self;
Surface::new<T: HasRawWindowHandle>(context: &Context, window: &T, width: u32, height: u32); 
Surface::resize(&mut self, width: u32, height: u32);
Surface::size(&self) -> (u32, u32);
Surface::buffer(&self) -> &[u8];
Surface::buffer_mut(&mut self) -> &mut [u8];
Surface::swap_buffers(&mut self);
```

(Perhaps make all of these return a `Result`.)

Considering the Wayland implementation, which is what I'm familiar with:
* `Context`/`Surface`/`resize`/`swap_buffers` generally mirrors Glutin and other libraries for hardware accelerated rendering. Just instead of OpenGL calls to draw, we have `.buffer_mut()`.
[...]

Issue #37, opened 2022-12-21
Split wrapper around "display" and "window" into different types


ids1024:
@i509vcb suggested this on Matrix. I don't know about other platforms, but for Wayland it could definitely be desirable to not create a new registry, etc. for every window. Doesn't make a difference for uses with a single window, but with a lot of `wl_surface`s it definitely is a good idea to not duplicate everything like that.

PR #46, opened 2022-12-23
x11: Implement image transfer using the MIT-SHM extension

notgull:
This PR implements image transfer using the MIT-SHM extension, allowing us to avoid image transfers over the wire.

~~At the moment I can't test this on my machine, as the x11-dl crate has a bug that prevents it from loading the MIT-SHM extension (https://github.com/AltF02/x11-rs/pull/164).~~ Circumvented this by moving to `x11rb`.

Closes https://github.com/rust-windowing/softbuffer/issues/38

(This is a port of rust-windowing/swbuf#23 to the new repository)

### DIFF
--- a/src/x11.rs
+++ b/src/x11.rs
@@ -14,7 +14,8 @@ use std::{fmt, io};
 use x11_dl::xlib::Display;
 use x11_dl::xlib_xcb::Xlib_xcb;
 
-use x11rb::connection::{Connection, RequestConnection};
+use x11rb::connection::{Connection, RequestConnection, SequenceNumber};
+use x11rb::cookie::Cookie;
 use x11rb::errors::{ConnectionError, ReplyError, ReplyOrIdError};
 use x11rb::protocol::shm::{self, ConnectionExt as _};
 use x11rb::protocol::xproto::{self, ConnectionExt as _};
@@ -41,6 +42,20 @@ pub struct X11Impl {
 struct ShmInfo {
     /// The shared memory segment, paired with its ID.
     seg: Option<(ShmSegment, shm::Seg)>,
+
+    /// A cookie indicating that the shared memory segment is ready to be used.
+    ///
+    /// We can't soundly read from or write to the SHM segment until the X server is done processing the
+    /// `shm::PutImage` request. However, the X server handles requests in order, which means that, if
+    /// we send a very small request after the `shm::PutImage` request, then the X server will have to
+    /// process that request before it can process the `shm::PutImage` request. Therefore, we can use
+    /// the reply to that small request to determine when the `shm::PutImage` request is done.
+    ///
+    /// In this case, we use `GetInputFocus` since it is a very small request.
+    ///
+    /// We store the sequence number instead of the `Cookie` since we cannot hold a self-referential
+    /// reference to the `connection` field.
+    done_processing: Option<SequenceNumber>,
 }
 
 impl X11Impl {
@@ -143,7 +158,10 @@ impl X11Impl {
 
             if present {
                 // SHM is available.
-                Some(ShmInfo { seg: None })
+                Some(ShmInfo {
+                    seg: None,
+                    done_processing: None,
+                })
             } else {
                 None
             }
@@ -192,12 +210,15 @@ impl X11Impl {
             None => return Ok(false),
         };
 
+        // If the X server is still processing the last image, wait for it to finish.
+        shm_info.finish_wait(&self.connection)?;
+
         // Get the SHM segment to use.
         let necessary_size = (width as usize) * (height as usize) * 4;
         let (segment, segment_id) = shm_info.segment(&self.connection, necessary_size)?;
 
         // Copy the buffer into the segment.
-        // SAFETY: The buffer is properly sized.
+        // SAFETY: The buffer is properly sized and we've ensured that the X server isn't reading from it.
         unsafe {
             segment.copy(buffer);
         }
@@ -222,6 +243,9 @@ impl X11Impl {
                 0,
             )?
             .ignore_error();
+
+        // Send a short request to act as a notification for when the X server is done processing the image.
+        shm_info.begin_wait(&self.connection)?;
 
         Ok(true)
     }
@@ -299,6 +323,25 @@ impl ShmInfo {
 
         Ok(())
     }
+
+    /// Begin waiting for the SHM processing to finish.
+    fn begin_wait(&mut self, c: &impl Connection) -> Result<(), PushBufferError> {
+        let cookie = c.get_input_focus()?.sequence_number();
+        let old_cookie = self.done_processing.replace(cookie);
+        debug_assert!(old_cookie.is_none());
+        Ok(())
+    }
+
+    /// Wait for the SHM processing to finish.
+    fn finish_wait(&mut self, c: &impl Connection) -> Result<(), PushBufferError> {
+        if let Some(done_processing) = self.done_processing.take() {
+            // Cast to a cookie and wait on it.
+            let cookie = Cookie::<_, xproto::GetInputFocusReply>::new(c, done_processing);
+            cookie.reply()?;
+        }
+
+        Ok(())
+    }
 }
 
 struct ShmSegment {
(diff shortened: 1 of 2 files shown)


src/x11.rs
```diff
@@ -74,19 +105,96 @@ impl X11Impl {
...
+            xshm: xshm.map(|xshm| ShmExtension { xshm, shmseg: None }),
             gc,
             visual,
             depth,
         })
     }
 
     pub(crate) unsafe fn set_buffer(&mut self, buffer: &[u32], width: u16, height: u16) {
+        if let Err(e) = self.shm_set(buffer, width, height) {
+            eprintln!("XShm not available, falling back to XImage: {e}");
```
i509VCB, line 126: We should probably use the `log` crate instead of println

src/x11.rs
```diff
@@ -74,19 +105,96 @@ impl X11Impl {
...
+        let shmseg = match &mut shm_ext.shmseg {
+            None => shm_ext.shmseg.insert(ShmSegment::new(shmseg_size)?),
+            Some(ref shmseg) if shmseg.size < shmseg_size => {
+                shm_ext.shmseg.insert(ShmSegment::new(shmseg_size)?)
+            }
+            Some(shmseg) => shmseg,
+        };
+
+        // Create the basic image.
+        let mut seg: XShmSegmentInfo = mem::zeroed();
```
i509VCB, line 154: I'd think using MaybeUninit would be clearer

src/x11.rs
notgull: I would prefer using the `tracing` crate if possible.

ids1024: Looks like `winit` uses `log`. Might make sense to use the same thing?

This should probably be a debug message if it's completely expected for some X servers or a remote connection, and doesn't indicate any kind of error.

src/x11.rs
ids1024: I don't know if `mem::zeroed()` is unclear. Seems fine to meas long as all zeros is a valid bit pattern for this type (otherwise it would be unsound).

Using uninitialized memory could technically be faster, and still sound, but it may be safer to stick with either `mem::zeroed()` or `MaybeUninit::zeroed()`.

notgull: Once #52 is merged I'll rewrite this PR to use `x11rb`, which should circumvent this issue entirely.

notgull:
I'm waiting for #55 to make progress before I continue on this, since I'll probably reuse whatever SHM mechanism that uses here.

jackpot51:
> I'm waiting for #55 to make progress before I continue on this, since I'll probably reuse whatever SHM mechanism that uses here.

55 is merged

jackpot51:
Will this gracefully fall back to the old method if MIT-SHM is not available? Would that be the case when using X11 forwarding, like with SSH?

notgull:
Yes, this falls back to the wire method if it determines that SHM is not available.

This is still marked as draft because I realized that I left a leak in it before I left for the gym. Let me fix that before I mark it as ready.

src/x11.rs
```diff
@@ -131,33 +137,170 @@ impl X11Impl {
...
 
+        // See if SHM is available.
+        let shm_info = {
+            let present = query_token
+                .reply()
+                .swbuf_err("Failed to get SHM query reply")?
+                .present;
+
+            if present {
+                // SHM is available.
```
psychon, line 148: Well... no. This only checks whether the SHM *extension* is available. But this does not check whether the X11 server and this process run on the same machine.... I think.

Are you sure the X11 server hides the SHM extension for non-local clients? How would that even work with ssh X11 forwarding? The X11 server would not know that ssh is forwarding something non-local, would it?

Anyway, cairo does a lot more complicated things to figure out whether SHM is usable: https://gitlab.freedesktop.org/cairo/cairo/-/blob/13a0474641232cccb15d24fa31d8c2557731665b/src/cairo-xcb-connection.c#L443-479

This seems to set up some shared memory and actually check whether a SHM Attach and Detach request succeed.

src/x11.rs
```diff
@@ -131,33 +137,170 @@ impl X11Impl {
...
+        height: u16,
+    ) -> Result<bool, PushBufferError> {
+        let shm_info = match self.shm {
+            Some(ref mut info) => info,
+            None => return Ok(false),
+        };
+
+        // Get the SHM segment to use.
+        let necessary_size = (width as usize) * (height as usize) * 4;
+        let (segment, segment_id) = shm_info.segment(&self.connection, necessary_size)?;
```
psychon, line 200: Does anything make sure that the X11 server is done with reading from the shared memory before this writes to it again? (After all, this is the reason why `shm_put_image` can generate completion events via the `send_event` argument.)

src/x11.rs
```diff
@@ -190,6 +333,33 @@ impl ShmSegment {
             Ok(Self { id, ptr, size })
         }
     }
+
+    /// Copy data into this shared memory segment.
+    ///
+    /// # Safety
+    ///
+    /// This function assumes that the size of `self`'s buffer is larger than or equal to `data.len()`.
```
psychon, line 341: Would it hurt much to make this safe and add an `assert!` or something like that? That extra check should be cheap compared to the `memcpy` that follows.

src/x11.rs
notgull: I used something similar that sets up shared memory and then attaches/detaches a buffer. This is probably something that we should split into the display if/when we do #37

src/x11.rs
notgull: I originally had `send_event` sent to `true` and then used a loop that would call `wait_for_event` until it received a valid SHM completion event. However, I got paranoid that this loop would pick up and discard an actual valid input event that [`winit`](https://crates.io/crates/winit) might be concerned about.

Besides, [it doesn't look like `cairo` does it](https://github.com/freedesktop/cairo/blob/12cd2bcbb260f4ed64a72b64bd2fdd593f683bbc/src/cairo-xcb-connection-shm.c#L69-L71), although there might be some ameliorating control I missed while browsing `cairo`'s source code.

psychon: Cairo does not control the main loop and thus cannot wait/poll for events. Instead, it (ab)uses the fact the X11 server handles requests in-order: It [sends a GetInputFocus request when it needs some syncronisation](https://github.com/freedesktop/cairo/blob/12cd2bcbb260f4ed64a72b64bd2fdd593f683bbc/src/cairo-xcb-shm.c#L303)  and then [uses `xcb_poll_for_reply()` to check whether that request is done](https://github.com/freedesktop/cairo/blob/12cd2bcbb260f4ed64a72b64bd2fdd593f683bbc/src/cairo-xcb-shm.c#L117).

notgull: That sounds like a good strategy, I've implemented it. Besides, I believe it's also consistent with the fact that the Wayland backend blocks until the buffer's ready.

ids1024:
I guess on vaguely modern versions of X11, we have the option to use https://docs.rs/x11rb/0.11.0/x11rb/protocol/shm/trait.ConnectionExt.html#method.shm_attach_fd instead of using `shm_attach` with System V shared memory segments? Not sure if there are an non-historical context in which shm would be available but that isn't. Or if there are any other drawbacks.

That could share code for allocating/growing with Wayland.

notgull:
> I guess on vaguely modern versions of X11, we have the option to use https://docs.rs/x11rb/0.11.0/x11rb/protocol/shm/trait.ConnectionExt.html#method.shm_attach_fd instead of using `shm_attach` with System V shared memory segments? Not sure if there are an non-historical context in which shm would be available but that isn't. Or if there are any other drawbacks.

The main drawback would be not being able to use `IPC_PRIVATE`, although I think you mentioned something earlier about `IPC_PRIVATE` causing issues on BSD? If so I can rewrite this to use `shm_open` et al.

notgull:
If it's all the same, then I'd rather keep the idiomatic SHM wrapper I have right now.

jackpot51:
I think this can be merged once @ids1024 approves

ids1024:
The code generally looks good now. Has it been tested that it correctly falls back to not using shm on a remote X server?

With XShm, any other X client could get the shm segment for our window, map it, and read our write to our window buffer, right? This is a bit of a concern for https://github.com/rust-windowing/softbuffer/issues/40. I think(?) that shouldn't be a soundness concern with `mem::copy_nonoverlapping` as used here.

notgull:
> The code generally looks good now. Has it been tested that it correctly falls back to not using shm on a remote X server?

I've tested it, and it successfully falls back to the wire method for transfer, so we're good there.

> With XShm, any other X client could get the shm segment for our window, map it, and read or write to our window buffer, right?

I don't believe that there is a way for an X client to get the SHM ID in use by another X client unless there's a server bug.

> I think this can be merged once @ids1024 approves

Okay, I'll merge this PR!

ids1024:
Ah. I was thinking that a compositor/compositing WM would read directly from the shm buffer, but I guess that's handled in a different way.

PR #52, opened 2022-12-23
x11: Add XCB support to the X11 backend

### LINKED ISSUE #45: Supporting both Xlib and Xcb raw window handles
X11 is one of the rare platforms with two libraries that interface with the protocol. Ideally swbuf should support working with both Xcb and Xlib handles.

notgull:
This PR adds XCB support to the X11 backend. It uses the [`x11rb`](https://crates.io/crates/x11rb) crate to handle interfacing with `libxcb`. In addition, the drawing logic is reimplemented using `x11rb`. Xlib support is now implemented by getting the inner `libxcb` connection from the Xlib `Display` and then just wrapping that.

I chose `x11rb` because it's at arm's length, I trust it, it has minimal dependencies, and it provides easy `libxcb` support. I'm down to bikeshed if we should use another wrapper.

Closes #45

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,18 +15,21 @@ exclude = ["examples"]
 default = ["x11", "wayland", "wayland-dlopen"]
 wayland = ["wayland-backend", "wayland-client", "nix"]
 wayland-dlopen = ["wayland-sys/dlopen"]
-x11 = ["x11-dl"]
+x11 = ["bytemuck", "x11rb", "x11-dl"]
 
 [dependencies]
 thiserror = "1.0.30"
 raw-window-handle = "0.5.0"
+log = "0.4.17"
 
 [target.'cfg(any(target_os = "linux", target_os = "freebsd"))'.dependencies]
 nix = { version = "0.26.1", optional = true }
 wayland-backend = { version = "0.1.0-beta.14", features = ["client_system"], optional = true }
 wayland-client = { version = "0.30.0-beta.14", optional = true }
 wayland-sys = "0.30.0"
+bytemuck = { version = "1.12.3", optional = true }
 x11-dl = { version  = "2.19.1", optional = true }
+x11rb = { version = "0.11.0", features = ["allow-unsafe-code", "dl-libxcb"], optional = true }
 
 [target.'cfg(target_os = "windows")'.dependencies.windows-sys]
 version = "0.42.0"
--- a/examples/libxcb.rs
+++ b/examples/libxcb.rs
@@ -1,0 +1,136 @@
+//! Example of using `softbuffer` with `libxcb`.
+
+#[cfg(all(feature = "x11", any(target_os = "linux", target_os = "freebsd")))]
+mod example {
+    use raw_window_handle::{RawDisplayHandle, RawWindowHandle, XcbDisplayHandle, XcbWindowHandle};
+    use softbuffer::GraphicsContext;
+    use x11rb::{
+        connection::Connection,
+        protocol::{
+            xproto::{self, ConnectionExt as _},
+            Event,
+        },
+        xcb_ffi::XCBConnection,
+    };
+
+    pub(crate) fn run() {
+        // Create a new XCB connection
+        let (conn, screen) = XCBConnection::connect(None).expect("Failed to connect to X server");
+
+        // x11rb doesn't use raw-window-handle yet, so just create our own.
+        let mut display_handle = XcbDisplayHandle::empty();
+        display_handle.connection = conn.get_raw_xcb_connection() as *mut _;
+        display_handle.screen = screen as _;
+
+        // Create a new window.
+        let mut width = 640u16;
+        let mut height = 480u16;
+
+        let window = conn.generate_id().unwrap();
+        let screen = &conn.setup().roots[screen];
+        let (root_visual, root_parent) = (screen.root_visual, screen.root);
+        conn.create_window(
+            x11rb::COPY_FROM_PARENT as _,
+            window,
+            root_parent,
+            0,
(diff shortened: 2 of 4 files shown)


john01dav:
I may have more in depth comments later, but for right now I have a question. It looks like you aren't adding any new examples, so I'm wondering how this is tested. Merging the new platform without at least one example that can exercise that code seems non-ideal I think. I assume that you have something local for testing, so perhaps it could be cleaned up if needed and added as an example?

notgull:
Since the Xlib code is now just a wrapper around the XCB code, the tests that already exist should still apply to it. So I just ran the examples, made the changes, and then verified that the examples still work as expected.

If you want, I can add a specific example where I make a libxcb connection and then add it.

ids1024:
Yeah, it would be good to have some sort of test for it. It should work if the xlib version works and is just wrapping the xcb version, but things that "should" work have a tendency of breaking.

X11rb generally seems better than x11-dl. But a couple things:
* x11-dl uses `dlopen`, but this doesn't seem to. I guess this creates a hard runtime dependency on libxcb if build with the `x11` feature, in contrast to how it previously worked, how winit works, and how wayland works with the default `wayland-dlopen` feature?
  - Though at least for the immediate future, XWayland and libxcb is probably on any system that uses Wayland. Other than more unusual configurations where someone could just recompile without the `x11` feature.
* `winit` uses x11-dl too. Would x11rb be better to use in winit too?
  - Since softbuffer is most likely to be used with winit, it is preferable but not essential to only depend on one X client crate. And if x11rb is simply the better choice overall, that should be true for winit too.

notgull:
> Yeah, it would be good to have some sort of test for it. It should work if the xlib version works and is just wrapping the xcb version, but things that "should" work have a tendency of breaking.

Okay, I added an example that just uses pure libxcb.

>  x11-dl uses `dlopen`, but this doesn't seem to. I guess this creates a hard runtime dependency on libxcb if build with the `x11` feature, in contrast to how it previously worked, how winit works, and how wayland works with the default `wayland-dlopen` feature?

With the `dl-libxcb` feature (which I use here), `x11rb` [dynamically loads `libxcb` using the `libloading` crate](https://github.com/psychon/x11rb/blob/b8aa0f57f757f061faea6b48d9e09e21a93da76e/x11rb/src/xcb_ffi/raw_ffi/ffi.rs#L15). `libloading` uses `dlopen` under the hood, IIRC.

> `winit` uses x11-dl too. Would x11rb be better to use in winit too?

I'd very much like to do this, and if `winit` would have me I'd be willing to reimplement the X11 backend in `x11rb`.  `x11-dl`, as a crate, scares me. It has several soundness violations that aren't fixable without a breaking change.  I've generally avoided it until now due to this, and after a glance at its source code I can spot an [aliasing violation](https://github.com/AltF02/x11-rs/blob/032d968411cc7686949765c10f9fc3afa27bab41/x11-dl/src/link.rs#L47) (I'm not really sure why it uses gradual initialization in the first place here). The library it wraps around is so horribly broken and unworkable that its maintainers created `libxcb` as an alternative for it. The only reason to keep using it is IME support, but you can just use the `XlibXCB` trick to get around needed to use Xlib for everything, and even then [there's a crate for that nowadays](https://crates.io/crates/xim).

ids1024:
Ah, it's using libloading. I didn't look at features and grepped for `dlopen`. Since we have a default `wayland-dlopen` feature (copied from winit), maybe an `x11-dlopen` feature would be good to have for completeness.

> The library it wraps around is so horribly broken and unworkable that its maintainers created libxcb as an alternative for it.

Weren't some of the horribly broken things about xlib tied to the implementation and fixed when xlib was re-implemented on top of libxcb? But yeah, from my limited understanding of X, using xcb is a better choice.

It's probably overall good to use x11rb here regardless of what winit does, but I'd be interested if @kchibisov has any thoughts on this.

kchibisov:
>It's probably overall good to use x11rb here regardless of what winit does, but I'd be interested if @kchibisov has any thoughts on this.

It doesn't matter what to use in the end, though I do trust xcb more than x11rb, no offense. Though, please, don't use Xlib, since otherwise you'll need to integrate the error handling in a way glutin does, though winit exposes hook, it just looks wrong...

psychon:
> though I do trust xcb more than x11rb, no offense.

None taken, but I am still curious: Is this just a gut feeling (e.g. "xcb is more mature and used; x11rb is just a tiny new Rust library that basically no one uses") or is there some specific reason? Let me to a quick attempt to reverse that feeling: https://gitlab.freedesktop.org/xorg/lib/libxcb/-/issues/36 and https://gitlab.freedesktop.org/xorg/lib/libxcb/-/issues/60 and https://gitlab.freedesktop.org/xorg/lib/libxcb/-/issues/32 are bugs in the generated code in libxcb. x11rb is not affected by these problems.

(TL;DR: C `struct` padding working differently than the padding in the X11 protocol leading to b0rked parsing; some missing bound checks by design; some uninitialised bytes being sent to the X11 server in padding)

ids1024 (approved):
Seems to work as well as the previous implementation, and we seem to have agreed x11rb is a better option than x11-dl, at least.

Implementation looks fine, though I'm not an X expert.

ids1024:
Theoretically there's definitely some good reason to trust an established widely used library like xcb, but an implementation in Rust that largely avoids unsafe code also has a clear theoretical advantage. And seems like an entirely viable approach when both it and xcb generate a lot of their code from xml. So I like x11rb as a solution. In principle.

If there are any specific concerns with x11rb that may be different, though if they are solvable it would be great to address them, and switch to x11rb in winit.

PR #53, opened 2022-12-27
Update to stable wayland-rs 0.30, specify `rust-version`, rename `SwBufError` back to `SoftBufferError`

ids1024:
Wayland-rs 0.30 is released now. I also remembered we should add a `rust-version` to the manifest. And it seems this error name is the last thing that's left from the "swbuf" name.

### DIFF
--- a/src/x11.rs
+++ b/src/x11.rs
@@ -4,7 +4,7 @@
 //! drawn. A more effective implementation would use shared memory instead of the wire. In
 //! addition, we may also want to blit to a pixmap instead of a window.
 
-use crate::SwBufError;
+use crate::SoftBufferError;
 use raw_window_handle::{XcbDisplayHandle, XcbWindowHandle, XlibDisplayHandle, XlibWindowHandle};
 use std::fmt;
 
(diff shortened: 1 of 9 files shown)


src/x11.rs
```diff
@@ -39,15 +39,15 @@ impl X11Impl {
     pub unsafe fn from_xlib(
         window_handle: XlibWindowHandle,
         display_handle: XlibDisplayHandle,
-    ) -> Result<Self, SwBufError> {
+    ) -> Result<Self, SoftBufferError> {
         // TODO: We should cache the shared libraries.
 
         // Try to open the XlibXCB shared library.
-        let lib_xcb = Xlib_xcb::open().swbuf_err("Failed to open XlibXCB shared library")?;
+        let lib_xcb = Xlib_xcb::open().softbuffer_err("Failed to open XlibXCB shared library")?;
```
notgull, line 46: I find the `softbuffer_err` name a bit too verbose for my liking, do you mind changing it back to what it was before? This is just a convenience function that isn't publicly exposed.

notgull (commented):
LGTM aside from my comment

notgull (approved):

Issue #44, opened 2022-12-21
Android platform


i509VCB:
EDIT: A bit of discussion on this happened in https://github.com/rust-windowing/swbuf/pull/8#issuecomment-1362056178

MarijnS95:
The most likely way to implement this would be via `ANativeWindow_lock()`, whose API has been wrapped for the `ndk` crate at:

https://github.com/MarijnS95/ndk/compare/window-lock

We can merge this into the `ndk` crate after some slight cleanup.

ids1024:
Using `ANativeWindow_lock`, is there a way to request that the window uses an `BGRX` buffer like SoftBuffer currently assumes? Or would we have to deal with a different pixel format? (And if that could vary, things become even more complicated...)

ids1024:
Right. Hopefully we can at least expect `RGBX` to always work?

In comparison Wayland's `wl_shm` has a huge list of formats but only requires a compositor to support `argb8888` and `xrgb8888` (which a little endian, so the same as formats called BGRA/BGRX elsewhere).

It seems we can't expect to use the same pixel format everywhere while also getting performant no-copy presentation.

MarijnS95:
First draft pushed at https://github.com/MarijnS95/softbuffer/compare/android, untested but at least letting you folks know that this is being worked on :)

MarijnS95:
Updated the branch, it is fully working now (but will need some cleanup before it is ready for a PR, though let me know if I should open it as draft already)!

A couple important notes:

- Android only has `R8G8B8A8_UNORM` and `R8G8B8X8_UNORM` as 32-bits-per-pixel format;
- Android (and/or the hardware) really values alignment, so consecutive scanlines of pixels are not contiguous: instead there's a stride parameter in pixel count;
  - For now I've exposed this value on `Buffer::stride()`, though we could also have pixel setters with `x, y` coords or similar.
- The `winit` example needed a few tune-ups to follow proper `Resumed` semantics, I can PR that separately (as well as the README reformat);
- The example cannot compile as an executable and `cdylib` for Android concurrently. I should perhaps make a separate `winit_android.rs` for that using a shared `.rs` file with the business logic, or thanks to most of the CFG's in the code:  an `[[example]] path = ""` to compile the same file twice;
- https://github.com/rust-mobile/ndk/pull/404 needs a review, merge and release as a preliminary: `softbuffer` is a neat little test-bed.

ids1024:
[...]

PR #54, opened 2022-12-27
Add support and CI tests for BSDs

ids1024:
This adds a fallback using `shm_open`/`shm_unlink` for platforms where `memfd_create` doesn't exist. This seems to be how this is normally handled, though it's a bit ugly.

This also builds the wayland/x11 code for NetBSD/OpenBSD/DragonFlyBSD.

Add CI builds for FreeBSD and NetBSD. We would need some kind of virtualisation though to actually run tests on such targets.

I've tested the `shm_open` logic on Linux, but haven't run it on any BSDs.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,7 @@
+fn main() {
+    cfg_aliases::cfg_aliases! {
+        free_unix: { all(unix, not(any(target_vendor = "apple", target_os = "android", target_os = "redox"))) },
+        x11_platform: { all(feature = "x11", free_unix, not(target_arch = "wasm32")) },
+        wayland_platform: { all(feature = "wayland", free_unix, not(target_arch = "wasm32")) },
+    }
+}
--- a/src/wayland/buffer.rs
+++ b/src/wayland/buffer.rs
@@ -1,4 +1,3 @@
-use nix::sys::memfd::{memfd_create, MemFdCreateFlag};
 use std::{
     ffi::CStr,
     fs::File,
@@ -15,6 +14,50 @@ use wayland_client::{
 
 use super::State;
 
+#[cfg(any(target_os = "linux", target_os = "freebsd"))]
+fn create_memfile() -> File {
+    use nix::sys::memfd::{memfd_create, MemFdCreateFlag};
+
+    let name = unsafe { CStr::from_bytes_with_nul_unchecked("softbuffer\0".as_bytes()) };
+    let fd = memfd_create(name, MemFdCreateFlag::MFD_CLOEXEC)
+        .expect("Failed to create memfd to store buffer.");
+    unsafe { File::from_raw_fd(fd) }
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "freebsd")))]
+fn create_memfile() -> File {
+    use nix::{
+        errno::Errno,
+        fcntl::OFlag,
+        sys::{
+            mman::{shm_open, shm_unlink},
+            stat::Mode,
+        },
+    };
+    use std::iter;
+
+    for _ in 0..=4 {
+        let mut name = String::from("softbuffer-");
+        name.extend(iter::repeat_with(fastrand::alphanumeric).take(7));
+        name.push('\0');
+
+        let name = unsafe { CStr::from_bytes_with_nul_unchecked(name.as_bytes()) };
+        // `CLOEXEC` is implied with `shm_open`
+        let fd = shm_open(
+            name,
+            OFlag::O_RDWR | OFlag::O_CREAT | OFlag::O_EXCL,
+            Mode::S_IRWXU,
+        );
+        if fd != Err(Errno::EEXIST) {
+            let fd = fd.expect("Failed to create POSIX shm to store buffer.");
+            let _ = shm_unlink(name);
+            return unsafe { File::from_raw_fd(fd) };
+        }
+    }
+
+    panic!("Failed to generate non-existant shm name")
+}
+
 pub(super) struct WaylandBuffer {
     qh: QueueHandle<State>,
     tempfile: File,
@@ -28,10 +71,7 @@ pub(super) struct WaylandBuffer {
 
 impl WaylandBuffer {
     pub fn new(shm: &wl_shm::WlShm, width: i32, height: i32, qh: &QueueHandle<State>) -> Self {
-        let name = unsafe { CStr::from_bytes_with_nul_unchecked("softbuffer\0".as_bytes()) };
-        let tempfile_fd = memfd_create(name, MemFdCreateFlag::MFD_CLOEXEC)
-            .expect("Failed to create memfd to store buffer.");
-        let tempfile = unsafe { File::from_raw_fd(tempfile_fd) };
+        let tempfile = create_memfile();
         let pool_size = width * height * 4;
         let pool = shm.create_pool(tempfile.as_raw_fd(), pool_size, qh, ());
         let released = Arc::new(AtomicBool::new(true));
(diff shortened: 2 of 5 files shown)


notgull (commented):
Is there a reason why `cfg(all(unix, not(target_vendor = "apple")))` wouldn't work for this?

LGTM otherwise

ids1024:
I was thinking the same thing, but thought that could be a problem for Android if/when that's added (https://github.com/rust-windowing/softbuffer/issues/44). And I guess Haiku is `cfg(unix)`  but has its own windowing system and variant in `raw-window-handle`, if it's ever added.

Annoying that Rust doesn't have a way to define a config option once that could be used everywhere for less repetition.

notgull:
We could use [`cfg_aliases`](https://crates.io/crates/cfg_aliases), like the main `winit` does now.

ids1024:
Updated to `cfg_aliases`, which is nice though it would be better if `cargo ` offered something like that, and it could also be used for platform-specific dependencies.

It's using `all(unix, not(any(target_vendor = "apple", target_os = "android")))`, so I guess it should work on Solaris or such too.

src/wayland/buffer.rs
```diff
@@ -15,6 +14,49 @@ use wayland_client::{
...
+        sys::{
+            mman::{shm_open, shm_unlink},
+            stat::Mode,
+        },
+    };
+    use rand::distributions::{Alphanumeric, DistString};
+
+    for _ in 0..=4 {
+        let mut name = String::from("softbuffer-");
+        Alphanumeric.append_string(&mut rand::thread_rng(), &mut name, 7);
```
notgull, line 41: This feels overkill, why not have a `static AtomicU64` that we increment every time to get the new ID?

ids1024: The POSIX shm namespace is global to the system, so that would collide with another window/application using softbuffer.

notgull: Ah, I see. Still, using `rand` feels like overkill, especially given how many dependencies it comes with. Have you considered [`fastrand`](https://crates.io/crates/fastrand)?

Also, isn't there a way to create a "private" SHM segment through `IPC_PRIVATE`?

build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() {
+    cfg_aliases::cfg_aliases! {
+        free_unix: { all(unix, not(any(target_vendor = "apple", target_os = "android", target_os = "redox"))) },
+        x11_platform: { all(feature = "x11", free_unix, not(wasm)) },
```
notgull, line 4: `wasm` isn't defined. You'd need an addition `cfg` alias to the tune of:

```
wasm: { target_family = "wasm" }
```

src/wayland/buffer.rs
```diff
@@ -15,6 +14,49 @@ use wayland_client::{
 
 use super::State;
 
+#[cfg(any(target_os = "linux", target_os = "freebsd"))]
+fn create_memfile() -> File {
+    use nix::sys::memfd::{memfd_create, MemFdCreateFlag};
```
notgull, line 19: I generally prefer `rustix` to `nix` when available, but given that most of `softbuffer`'s downstream dependencies use `nix` anyways, I'm fine with this for now.

src/wayland/buffer.rs
ids1024: `IPC_PRIVATE` is a System V shared memory flag, not a POSIX shared memory flag, so it's different. We can't use System V shared memory segments since they don't have file descriptors and can't be resized.

FreeBSD supports `SHM_ANON` in place of `name` for this, but it doesn't look like that's supported on any other platform and FreeBSD now offers a `memfd_create` function compatible with the Linux one. (Though I'm not sure when that was added, so this could be a concern for older FreeBSD versions).

Using a random name and a loop with `shm_open` and `shm_unlink` seems to be the standard approach. With variations. I see GTK's Wayland backend uses `sprintf (name, "/gdk-wayland-%x", g_random_int ());`, and the loop doesn't have any limit on the number of times it iterates.

If `rand` is a heavy dependency though, something else could be used. But I think some method of generating a random name is necessary.

src/wayland/buffer.rs
ids1024: Yeah, I found out about `rustix` recently and it could be a better choice. Though nix is indeed likely to be a dependency of other things. (Though this only matters if they depend on the same version of nix, anyway.)

src/wayland/buffer.rs
notgull: It doesn't look like `rand` shows up anywhere else in the `winit` dependency tree, so I'd be hesitant to add it. Like I said, I'd rather use a lighter dependency like `fastrand`.

src/wayland/buffer.rs
notgull: At the moment, most of the Wayland crates depend on `nix`, as well as `x11rb`. For the time being we probably should just stick with `nix`.

ids1024: Rustix is also less complete. It looks like it doesn't currently have wrappers around `shm_open` and `shm_unlink`.

Of course the minimal solution is just to use the libc crate.

src/wayland/buffer.rs
ids1024: Switched to `fastrand`.

src/wayland/buffer.rs
```diff
@@ -15,6 +14,49 @@ use wayland_client::{
...
+
+    let name = unsafe { CStr::from_bytes_with_nul_unchecked("softbuffer\0".as_bytes()) };
+    let fd = memfd_create(name, MemFdCreateFlag::MFD_CLOEXEC)
```
i509VCB, line 22: We should fallback to shm files if creating a memfd failed, as we could get `ENOTSUPP` on older versions of Linux.

src/wayland/buffer.rs
```diff
@@ -15,6 +14,49 @@ use wayland_client::{
...
+
+    for _ in 0..=4 {
+        let mut name = String::from("softbuffer-");
+        name.extend(iter::repeat_with(fastrand::alphanumeric).take(7));
+        name.push('\0');
+
+        let name = unsafe { CStr::from_bytes_with_nul_unchecked(name.as_bytes()) };
+        let fd = shm_open(
+            name,
+            OFlag::O_RDWR | OFlag::O_CREAT | OFlag::O_EXCL,
```
i509VCB, line 47: We created the memfd with the `CLOEXEC` flag on Linux and BSD but not here?

src/wayland/buffer.rs
ids1024: I think it will fail to compile anyway if libc doesn't have the function, so this would only help with a libc version that has the function and a kernel that doesn't?

There's also FreeBSD's `SHM_ANON` argument to `shm_open` which is useful here on older versions without `memfd_create`. But not useful if it doesn't compile anyway.

src/wayland/buffer.rs
ids1024: I wasn't thinking of this, but apparently that's implied?

>  The FD_CLOEXEC file descriptor flag associated with the new file descriptor is set.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/shm_open.html

i509VCB: Hmm okay, I guess we should be explicit about the flag though.

ids1024: I assume passing `O_CLOEXEC` wouldn't cause issues (and would just be ignored), but it's not listed here as one of the flags that can be passed.

i509VCB: Okay, I guess you could just mention the implicit flag in a doc comment then.

src/wayland/buffer.rs
ids1024: Including the fallback on Linux/FreeBSD also means depending on fastrand there. Which isn't a big deal, but seems unnecessary if it's going to just fail to compile where this code path would be used.

jackpot51: I agree that if it links to the function it is pribably implemented, either by the kernel or as a shim.

jackpot51 (approved):
Looks good to me.

Issue #48, opened 2022-12-23
Wayland animation example fails on Gnome


john01dav:
When running the animation example on Gnome/Mutter 43.2 it causes the entire desktop to freeze for about half a second, then the example terminates with the following error:
```
request could not be marshaled: can't send file descriptor
Error sending request: Resource temporarily unavailable
```

I've known about this bug for some time, but when I tried diving into Wayland to fix it, I was unable to find a solution. Maybe you can shed some light @ids1024?

Additionally, some other examples (notably winit_wrong_sized_buffer and fruit) are janky insofar as resizing causes temporary freezes of the softbuffer window.

i509VCB:
I assume winit has a way to indicate if a window has stopped resizing. We could present a single color there to make resizes much faster.

ids1024:
I've just been testing with the `winit` example. In cdfae58510895770c61ec3487b599d39aaf5338d I added code to get it to redraw on resize, which seems to be the same issue as `fruit` and `winit_wrong_sized_buffer`. I'm not sure if there's a more correct way to deal with that in winit.

As for the issue with `animation`, I don't see this with cosmic-comp as compositor.

ids1024:
Okay, I can reproduce the freeze and error on Gnome. The error occurs with smithay/cosmic-comp, but less consistently and without freezing the session.

It seems with `window.request_redraw()` called in `MainEventsCleared`, it ends up attaching buffers before the buffer is released? Which the code assumes should normally not happen, and just destroys the old buffer and pool and creates a new one.

So the way I implemented that doesn't seem to be adequate.

ids1024:
We could wait for a buffer release as suggested in https://github.com/rust-windowing/softbuffer/issues/41. But then it needs to be documented that `set_buffer` can block that way. Which is not always desirable, especially if your thread is doing things other than just drawing a UI.

ids1024:
For something like the animation example, anyway, the correct way to handle it would normally be to queue a redraw in response to the Wayland surface's frame callback, right? While it doesn't look like winit's Wayland backend makes use of frame callbacks?

While queuing a redraw on `MainEventsCleared` can send buffers faster than the compositor can display or release them if the drawing code is fast (and in `animation`, it is, since it pre-renders frames).

ids1024:
This is fixed if we change `set_buffer` to block not just for buffer release, but until a `frame` callback is called. (I guess this is basically "vsync"?)

[...]

Issue #41, opened 2022-12-21
Should Wayland allow double buffering or just wait until the buffer is released.


i509VCB:
On Wayland, we need to wait for the server to release a wl_buffer before we can reuse it. This generally means clients will double buffer and track buffer age when rendering. However it would be most ideal to avoid double buffering if at all possible.

This probably means that softbuffer also needs a way to check if the buffer can be acquired to be rendered to if we run out of buffers to use.

ids1024:
I guess an alternative to blocking would be to just drop the frame (have `set_buffer` do nothing if the back buffer hasn't been released)? But that could only work if the caller expects it to work that way, or it might not set a new buffer any time soon.

Maybe blocking until buffer release is best. While if the application were to drive its rendering based on a `frame` callback, this has no impact since the buffer would already be released.

jackpot51:
I believe this is fixed by #55

PR #55, opened 2022-12-27
wayland: Block dispatching if back buffer isn't released

ids1024:
https://github.com/rust-windowing/softbuffer/issues/41

If `set_buffer` can just be called in a loop without waiting for buffers to be released or a frame callback, but can also be called in other ways, I don't know if there's a better solution than blocking.

Should fix https://github.com/rust-windowing/softbuffer/issues/48. The animation example could probably be implemented better, but this is at least better.

I guess it should be documented that `set_buffer` may block? I don't know how this compares to other backends.

Leaving as a draft at the moment since https://github.com/rust-windowing/softbuffer/issues/48 still seems to be an issue. The error doesn't occur, but at least in a VM I'm seeing Gnome Shell freezing (not an issue with smithay/cosmic-comp). I guess it's still sending buffers to the server, without an FPS limit or use of the frame callback, and the server doesn't have time to do much else?

### DIFF
--- a/src/wayland/mod.rs
+++ b/src/wayland/mod.rs
@@ -1,6 +1,5 @@
 use crate::{error::unwrap, SoftBufferError};
 use raw_window_handle::{WaylandDisplayHandle, WaylandWindowHandle};
-use std::collections::VecDeque;
 use wayland_client::{
     backend::{Backend, ObjectId},
     globals::{registry_queue_init, GlobalListContents},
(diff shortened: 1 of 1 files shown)


ids1024:
Oh, I forgot about this part:

> A server should avoid signaling the frame callbacks if the surface is not visible in any way, e.g. the surface is off-screen, or completely obscured by other opaque surfaces.

https://wayland.app/protocols/wayland#wl_surface:request:frame

We definitely don't want to block indefinitely if a window isn't visible, so we can't wait on a frame callback, and really need the application/winit to handle that. (Even without that, this would be a problem with multiple windows).

So I guess this is the best we can do in softbuffer, at present.

ids1024:
@i509VCB Does this seem like a reasonable way to deal with double buffering? As you mentioned in https://github.com/rust-windowing/softbuffer/issues/41, double-buffering may not be necessary on compositors that just upload the data to GPU and don't keep the shm buffer, but I don't see an Wayland API that indicates if the compositor will do that. So if we tried to only allocate a second buffer only when the first isn't released, it may tend to eventually end up allocating a second unless the application waits for a frame callback to render.

jackpot51 (approved):
This looks reasonable to me

PR #57, opened 2023-01-01
Rework documentation for `GraphicsContext::set_buffer`

i509VCB:
This includes a more clear description of the pixel format expected.


src/lib.rs
```diff
@@ -160,28 +160,49 @@ impl GraphicsContext {
...
     ///
+    /// ```text
     /// 00000000RRRRRRRRGGGGGGGGBBBBBBBB
+    /// ^                              ^
+    /// MSB                          LSB
+    /// ```
+    ///
+    /// | Bit  | Channel       |
+    /// |------|---------------|
+    /// | 0    | Bit is `0`    |
```
ids1024, line 190: Isn't this `X` (the value of these bits is arbitrary and will be ignored) rather than `0`? Which isn't quite the same thing.

notgull (approved):

src/lib.rs
```diff
@@ -160,28 +160,49 @@ impl GraphicsContext {
...
+    ///
+    /// The function shown below may be used to create a color from a 24-bit RGB color using bitwise\
+    /// operations:
     ///
-    /// 0: Bit is 0
-    /// R: Red channel
-    /// G: Green channel
-    /// B: Blue channel
+    /// ```
+    /// fn pixel(red: u8, green: u8, blue: u8) -> u32 {
```
john01dav, line 199: Excellent addition to have this pre-made example function. I'm not sure if this is the right place for it. Perhaps it should be in an example instead, or just callable as a pub fn in softbuffer. I think that this is ultimately the best place for it, but I'm not certain so I wanted to call attention to this choice.

john01dav (changes requested):

src/lib.rs
john01dav: The existing code on every backend should ignore them, but that isn't the same as specifying they the correct use of this library is zero.

ids1024: Maybe there could be slightly more specific wording indicating that values other than 0 may have issues with certain backends?

src/lib.rs
i509VCB: The utility of this function being publicly accessible is probably quite low as you'd want a more optimal way to do this really.

i509VCB:
Wait what, why does github only allow me to request one reviewer but not multiple...

notgull:
@i509VCB I will accept this PR if it is rebased and rewritten to apply to the `Buffer` struct instead.

i509VCB:
Unfortunately do not have time to continue this.

PR #59, opened 2023-01-05
Disable CALayer fade action when setting buffers

thedjinn:
This pull request fixes a bug in the Core Graphics backend that causes a new buffer not be shown immediately but instead use a quarter second fade transition. This happens because the CALayer has a default action associated with a change in the layer contents.

The problem was mitigated by wrapping the contents change in a transaction and disabling all actions for the duration of this transaction.

A new example was added to demonstrate the problem and prove that the solution works.


jackpot51 (approved):
I was able to replicate the problem using the rectangle example and verify that this PR fixes it.

PR #64, opened 2023-01-06
Split display/window wrappers

ids1024:
WIP implementation of https://github.com/rust-windowing/softbuffer/issues/37. Currently compiles on Linux, and splits parts of Wayland code to not need duplication for multiple windows.

Seems best to start with this API change before moving on to https://github.com/rust-windowing/softbuffer/issues/40.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,6 +21,9 @@ mod x11;
 
 mod error;
 
+#[cfg(any(wayland_platform, x11_platform))]
+use std::sync::Arc;
+
 pub use error::SoftBufferError;
 
 use raw_window_handle::{
@@ -29,12 +32,9 @@ use raw_window_handle::{
 
 /// An instance of this struct contains the platform-specific data that must be managed in order to
 /// write to a window on that platform.
-pub struct GraphicsContext {
+pub struct Context {
     /// The inner static dispatch object.
-    ///
-    /// This is boxed so that `GraphicsContext` is the same size on every platform, which should
-    /// hopefully prevent surprises.
-    graphics_context_impl: Box<Dispatch>,
+    context_impl: ContextDispatch,
 }
 
 /// A macro for creating the enum used to statically dispatch to the platform-specific implementation.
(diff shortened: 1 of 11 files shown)


notgull:
`Context` is expected to be created once per connection, which brings up how often `Surface`s are expected to be created. Should it be created once per surface, kept around for the lifetime of that surface, and dropped only when the surface is dropped? Or would it be perfectly cheap to create and drop `Surface`s, like so:

```rust
match event {
    Event::RedrawRequested => {
        let surface = Surface::new(&context, my_window);
        /* .. */
        drop(surface);
    }
}
```

If the latter, we should move some of the allocations out of `Surface` into the `Context`. In either case, this should be documented.

src/lib.rs
```diff
@@ -29,30 +32,48 @@ use raw_window_handle::{
...
 /// write to a window on that platform.
-pub struct GraphicsContext {
+pub struct Context {
     /// The inner static dispatch object.
     ///
-    /// This is boxed so that `GraphicsContext` is the same size on every platform, which should
+    /// This is boxed so that `Context` is the same size on every platform, which should
     /// hopefully prevent surprises.
-    graphics_context_impl: Box<Dispatch>,
+    context_impl: Box<ContextDispatch>,
```
notgull, line 40: Since all of the variants here are either ZSTs or `Arc`s I'm fine with not boxing it.

ids1024:
I'm assuming one would keep the `Surface` and continue using it. So an allocations that are specific to the surface would be there.

src/lib.rs
ids1024: Ah right, I meant to remove the `Box` when I added the use of `Arc`s.

notgull (commented):
I'm wondering if it would be more idiomatic to make `set_buffer` take `&Context` as an argument. This would mean that we wouldn't need to have an `Arc<Context>` for Wayland and X11. This is less of a performance thing and more of an "is this more idiomatic?" thing.

LGTM otherwise

ids1024:
Maybe. Though taking a `&Context` could be a bit more inconvenient after https://github.com/rust-windowing/softbuffer/issues/40 since that will add more methods to `Surface`. And then you either have to pass it to all of them, or it's necessary to determine which methods might need something from the context (on some backends).

There could also be weird behavior if you passed a different context, though that probably isn't too much of an issue.

(Perhaps the ideal solution would be a lifetime bound on `Surface` if Rust had some kind of magical solution to make self-referential structs convenient, but that doesn't exist and makes lifetime bounds often awkward to deal with.)

Issue #17, opened 2022-06-14
Transparency


LoganDark:
Right now, softbuffer seems to completely ignore the alpha channel. It would be nice if this was supported.

i509VCB:
A transparency hint would be useful in the Wayland backend, as we could tell the compositor about the window content being opaque and let the compositor apply some optimizations for rendering.

ids1024:
In addition to a way to set this, it would be good to have an API to test if it's supported. On Wayland RGBA and RGBX should always be, but I think this may not be true on others?

What kind of API would be suitable here? A `set_alpha()` method? `supports_alpha()`? Maybe there's a better way to name it.

This becomes a lot more complicated if we wanted to support formats other than RGBA/RGBX.

cafce25:
It's probably by accident, but on windows 11 tranparency does work by writing the alpha value in the first byte which `Surface` specifies to always be 0:
```
let mut buffer = self.surface.buffer_mut()?;
let image_data = pixmap.data();
for index in 0..(width * height) as usize {
    let [r, g, b, a] = image_data[index * 4..][..4].try_into()?;
    buffer[index] = u32::from_be_bytes([a, r, g, b]);
}
```
It might be worth to preserve that behaviour if feasible.

madsmtm:
On macOS/iOS, I think it's possible to query the transparency state of the surface, something like [`-[CALayer isOpaque]`](https://developer.apple.com/documentation/quartzcore/calayer/isopaque?language=objc)? If that's possible on other platforms too, maybe we could make this entirely transparent (haha) to the user?

If not, then I'd be in favour of a `fn set_transparent(&mut self, state: bool)` on `Surface`, which switches between `BGRX` and `BGRA` (or whatever pixel format we end up with).

MarijnS95:
Android has [`ASurfaceTransaction_setBufferTransparency()`](https://developer.android.com/ndk/reference/group/native-activity#asurfacetransaction_setbuffertransparency) and [`ASurfaceTransaction_setBufferAlpha()`](https://developer.android.com/ndk/reference/group/native-activity#asurfacetransaction_setbufferalpha) which becomes relevant on my impending `ASurfaceControl`/`ASurfaceTransaction`-based Android backend for `softbuffer`.

[...]

Issue #13, opened 2022-06-11
buffer width and height are only u16


LoganDark:
This limits the window to 65,536 * 65,536. :/

Is there any reason for this limitation, does a backend mandate it or something?

i509VCB:
I know X11 limits window dimensions to a u16. Wayland is everything greater than zero within an `i32` (so effectively a u31).

LoganDark:
> I know X11 limits window dimensions to a u16.

This is a platform-specific-panic sort of situation - if you ever do encounter a window bigger than that, anyway.

kchibisov:
In what setup would you need to create a buffer that large? I'm not even sure you can achieve a resolution like that, and both windows and X11 limit to u16?

65536 * 65536 window will consume 128 GB of memory, sure, you'll have a window that large.

Not even speaking about bandwidth you'd need for it.

LoganDark:
> In what setup would you need to create a buffer that large? I'm not even sure you can achieve a resolution like that, and both windows and X11 limit to u16?
> 
> 65536 * 65536 window will consume 128 GB of memory, sure, you'll have a window that large.
> 
> Not even speaking about bandwidth you'd need for it.

You only have to exceed it in 1 dimension for there to be an issue, and 8K monitors famously have like 7680 pixels across, it's feasible to exceed 65536 *today* with like 10 of these monitors in a row.

This is somewhat of a contrived scenario but 65536 is only like 1 or 2 orders of magnitude off from a normal desktop setup which makes me uncomfortable. I would much rather softbuffer use u32 internally even if 80% of its backends will panic. Because then maybe those backends can be updated without breaking changes if the API is future-proof.

kchibisov:
>8–9 of these monitors in a row.

128GB of memory for a single frame, sure you have that many video memory for a single window on a GPU. maybe in 50 years. Like what you describe isn't possible and no system will ever handle that in 2 decades or unless we have RAM in TB instead of GB. So with 32TB of ram I'll be comfortable using such thing.

>Because then maybe those backends can be updated without breaking changes if the API is future-proof.

Breaking change to WinAPI/X11 protocol btw.

ids1024:
> it's feasible to exceed 65536 today with like 8–9 of these monitors in a row

"Feasible" may be an exaggeration. I don't think you'll find any GPU that can drive 9 8k displays. Can they even handle 2? If so that would require displays connected to 5 different graphics cards. At best "technically possible" rather than "feasible".

[...]

PR #65, opened 2023-01-11
Owned pixel buffer for no-copy presentation

ids1024:
WIP implementation of https://github.com/rust-windowing/softbuffer/issues/40.

Currently this only has Wayland. It can be run with `cargo run --no-default-features --features wayland --example winit`.

We should do some testing to see what impact it actually has.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ For now, the priority for new platforms is:
 Example
 ==
 ```rust,no_run
+use std::num::NonZeroU32;
 use winit::event::{Event, WindowEvent};
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::WindowBuilder;
@@ -72,7 +73,12 @@ fn main() {
                     let size = window.inner_size();
                     (size.width, size.height)
                 };
-                surface.resize(width, height).unwrap();
+                surface
+                    .resize(
+                        NonZeroU32::new(width).unwrap(),
+                        NonZeroU32::new(height).unwrap(),
+                    )
+                    .unwrap();
 
                 let mut buffer = surface.buffer_mut().unwrap();
                 for index in 0..(width * height) {
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ fn main() {
                 };
                 surface.resize(width, height).unwrap();
 
-                let buffer = surface.buffer_mut().unwrap();
+                let mut buffer = surface.buffer_mut().unwrap();
                 for index in 0..(width * height) {
                     let y = index / width;
                     let x = index % width;
@@ -85,7 +85,7 @@ fn main() {
                     buffer[index as usize] = blue | (green << 8) | (red << 16);
                 }
 
-		surface.present().unwrap();
+		buffer.present().unwrap();
             }
             Event::WindowEvent {
                 event: WindowEvent::CloseRequested,
--- a/README.md
+++ b/README.md
@@ -72,21 +72,20 @@ fn main() {
                     let size = window.inner_size();
                     (size.width, size.height)
                 };
-                let buffer = (0..((width * height) as usize))
-                    .map(|index| {
-                        let y = index / (width as usize);
-                        let x = index % (width as usize);
-                        let red = x % 255;
-                        let green = y % 255;
-                        let blue = (x * y) % 255;
+                surface.resize(width, height);
 
-                        let color = blue | (green << 8) | (red << 16);
+                let buffer = surface.buffer_mut();
+                for index in 0..(width * height) {
+                    let y = index / width;
+                    let x = index % width;
+                    let red = x % 255;
+                    let green = y % 255;
+                    let blue = (x * y) % 255;
 
-                        color as u32
-                    })
-                    .collect::<Vec<_>>();
+                    buffer[index as usize] = blue | (green << 8) | (red << 16);
+                }
 
-                surface.set_buffer(&buffer, width as u16, height as u16);
+		surface.present();
             }
             Event::WindowEvent {
                 event: WindowEvent::CloseRequested,
--- a/README.md
+++ b/README.md
@@ -72,9 +72,9 @@ fn main() {
                     let size = window.inner_size();
                     (size.width, size.height)
                 };
-                surface.resize(width, height);
+                surface.resize(width, height).unwrap();
 
-                let buffer = surface.buffer_mut();
+                let buffer = surface.buffer_mut().unwrap();
                 for index in 0..(width * height) {
                     let y = index / width;
                     let x = index % width;
@@ -85,7 +85,7 @@ fn main() {
                     buffer[index as usize] = blue | (green << 8) | (red << 16);
                 }
 
-		surface.present();
+		surface.present().unwrap();
             }
             Event::WindowEvent {
                 event: WindowEvent::CloseRequested,
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -26,7 +26,6 @@ pub struct CGImpl {
     layer: CALayer,
     window: id,
     color_space: CGColorSpace,
-    buffer: Option<Vec<u32>>,
     width: u32,
     height: u32,
 }
@@ -54,7 +53,6 @@ impl CGImpl {
             color_space,
             width: 0,
             height: 0,
-            buffer: None,
         })
     }
 
@@ -64,45 +62,57 @@ impl CGImpl {
         Ok(())
     }
 
-    pub fn buffer_mut(&mut self) -> Result<&mut [u32], SoftBufferError> {
-        if self.buffer.is_none() {
-            self.buffer = Some(Vec::new());
-        }
-        let buffer = self.buffer.as_mut().unwrap();
-        buffer.resize(self.width as usize * self.height as usize, 0);
-        Ok(buffer.as_mut())
+    pub fn buffer_mut(&mut self) -> Result<BufferImpl, SoftBufferError> {
+        Ok(BufferImpl {
+            buffer: vec![0; self.width as usize * self.height as usize],
+            imp: self,
+        })
+    }
+}
+
+pub struct BufferImpl<'a> {
+    imp: &'a mut CGImpl,
+    buffer: Vec<u32>,
+}
+
+impl<'a> BufferImpl<'a> {
+    pub fn pixels(&self) -> &[u32] {
+        &self.buffer
     }
 
-    pub fn present(&mut self) -> Result<(), SoftBufferError> {
-        if let Some(buffer) = self.buffer.take() {
-            let data_provider = CGDataProvider::from_buffer(Arc::new(Buffer(buffer)));
-            let image = CGImage::new(
-                self.width as usize,
-                self.height as usize,
-                8,
-                32,
-                (self.width * 4) as usize,
-                &self.color_space,
-                kCGBitmapByteOrder32Little | kCGImageAlphaNoneSkipFirst,
-                &data_provider,
-                false,
-                kCGRenderingIntentDefault,
-            );
+    pub fn pixels_mut(&mut self) -> &mut [u32] {
+        &mut self.buffer
+    }
 
-            // The CALayer has a default action associated with a change in the layer contents, causing
-            // a quarter second fade transition to happen every time a new buffer is applied. This can
-            // be mitigated by wrapping the operation in a transaction and disabling all actions.
-            transaction::begin();
-            transaction::set_disable_actions(true);
+    pub fn present(self) -> Result<(), SoftBufferError> {
+        let data_provider = CGDataProvider::from_buffer(Arc::new(Buffer(self.buffer)));
+        let image = CGImage::new(
+            self.imp.width as usize,
+            self.imp.height as usize,
+            8,
+            32,
+            (self.imp.width * 4) as usize,
+            &self.imp.color_space,
+            kCGBitmapByteOrder32Little | kCGImageAlphaNoneSkipFirst,
+            &data_provider,
+            false,
+            kCGRenderingIntentDefault,
+        );
 
-            unsafe {
-                self.layer
-                    .set_contents_scale(self.window.backingScaleFactor());
-                self.layer.set_contents(image.as_ptr() as id);
-            };
+        // The CALayer has a default action associated with a change in the layer contents, causing
+        // a quarter second fade transition to happen every time a new buffer is applied. This can
+        // be mitigated by wrapping the operation in a transaction and disabling all actions.
+        transaction::begin();
+        transaction::set_disable_actions(true);
 
-            transaction::commit();
-        }
+        unsafe {
+            self.imp
+                .layer
+                .set_contents_scale(self.imp.window.backingScaleFactor());
+            self.imp.layer.set_contents(image.as_ptr() as id);
+        };
+
+        transaction::commit();
 
         Ok(())
     }
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -12,6 +12,7 @@ use cocoa::base::{id, nil};
 use cocoa::quartzcore::{transaction, CALayer, ContentsGravity};
 use foreign_types::ForeignType;
 
+use std::num::NonZeroU32;
 use std::sync::Arc;
 
 struct Buffer(Vec<u32>);
@@ -56,9 +57,9 @@ impl CGImpl {
         })
     }
 
-    pub fn resize(&mut self, width: u32, height: u32) -> Result<(), SoftBufferError> {
-        self.width = width;
-        self.height = height;
+    pub fn resize(&mut self, width: NonZeroU32, height: NonZeroU32) -> Result<(), SoftBufferError> {
+        self.width = width.get();
+        self.height = height.get();
         Ok(())
     }
 
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -103,12 +103,6 @@ impl CGImpl {
             transaction::commit();
         }
     }
-
-    pub(crate) unsafe fn set_buffer(&mut self, buffer: &[u32], width: u16, height: u16) {
-        self.resize(width.into(), height.into());
-        self.buffer_mut().copy_from_slice(buffer);
-        self.present();
-    }
 }
 
 impl Drop for CGImpl {
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -76,10 +76,12 @@ pub struct BufferImpl<'a> {
 }
 
 impl<'a> BufferImpl<'a> {
+    #[inline]
     pub fn pixels(&self) -> &[u32] {
         &self.buffer
     }
 
+    #[inline]
     pub fn pixels_mut(&mut self) -> &mut [u32] {
         &mut self.buffer
     }
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -68,7 +68,7 @@ impl CGImpl {
             self.buffer = Some(Vec::new());
         }
         let buffer = self.buffer.as_mut().unwrap();
-        buffer.resize(self.width as usize * self.height as usize * 4, 0);
+        buffer.resize(self.width as usize * self.height as usize, 0);
         buffer.as_mut()
     }
 
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -58,18 +58,19 @@ impl CGImpl {
         })
     }
 
-    pub fn resize(&mut self, width: u32, height: u32) {
+    pub fn resize(&mut self, width: u32, height: u32) -> Result<(), SoftBufferError> {
         self.width = width;
         self.height = height;
+        Ok(())
     }
 
-    pub fn buffer_mut(&mut self) -> &mut [u32] {
+    pub fn buffer_mut(&mut self) -> Result<&mut [u32], SoftBufferError> {
         if self.buffer.is_none() {
             self.buffer = Some(Vec::new());
         }
         let buffer = self.buffer.as_mut().unwrap();
         buffer.resize(self.width as usize * self.height as usize, 0);
-        buffer.as_mut()
+        Ok(buffer.as_mut())
     }
 
     pub fn present(&mut self) -> Result<(), SoftBufferError> {
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -72,7 +72,7 @@ impl CGImpl {
         buffer.as_mut()
     }
 
-    pub fn present(&mut self) {
+    pub fn present(&mut self) -> Result<(), SoftBufferError> {
         if let Some(buffer) = self.buffer.take() {
             let data_provider = CGDataProvider::from_buffer(Arc::new(Buffer(buffer)));
             let image = CGImage::new(
@@ -102,6 +102,8 @@ impl CGImpl {
 
             transaction::commit();
         }
+
+        Ok(())
     }
 }
 
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -14,8 +14,20 @@ use foreign_types::ForeignType;
 
 use std::sync::Arc;
 
+struct Buffer(Vec<u32>);
+
+impl AsRef<[u8]> for Buffer {
+    fn as_ref(&self) -> &[u8] {
+        bytemuck::cast_slice(&self.0)
+    }
+}
+
 pub struct CGImpl {
     layer: CALayer,
+    color_space: CGColorSpace,
+    buffer: Option<Vec<u32>>,
+    width: u32,
+    height: u32,
 }
 
 impl CGImpl {
@@ -34,36 +46,61 @@ impl CGImpl {
             view.addSubview_(subview); // retains subview (+1) = 2
             let _: () = msg_send![subview, release]; // releases subview (-1) = 1
         }
-        Ok(Self { layer })
+        let color_space = CGColorSpace::create_device_rgb();
+        Ok(Self {
+            layer,
+            color_space,
+            width: 0,
+            height: 0,
+            buffer: None,
+        })
+    }
+
+    pub fn resize(&mut self, width: u32, height: u32) {
+        self.width = width;
+        self.height = height;
+    }
+
+    pub fn buffer_mut(&mut self) -> &mut [u32] {
+        if self.buffer.is_none() {
+            self.buffer = Some(Vec::new());
+        }
+        let buffer = self.buffer.as_mut().unwrap();
+        buffer.resize(self.width as usize * self.height as usize * 4, 0);
+        buffer.as_mut()
+    }
+
+    pub fn present(&mut self) {
+        if let Some(buffer) = self.buffer.take() {
+            let data_provider = CGDataProvider::from_buffer(Arc::new(Buffer(buffer)));
+            let image = CGImage::new(
+                self.width as usize,
+                self.height as usize,
+                8,
+                32,
+                (self.width * 4) as usize,
+                &self.color_space,
+                kCGBitmapByteOrder32Little | kCGImageAlphaNoneSkipFirst,
+                &data_provider,
+                false,
+                kCGRenderingIntentDefault,
+            );
+
+            // The CALayer has a default action associated with a change in the layer contents, causing
+            // a quarter second fade transition to happen every time a new buffer is applied. This can
+            // be mitigated by wrapping the operation in a transaction and disabling all actions.
+            transaction::begin();
+            transaction::set_disable_actions(true);
+
+            unsafe { self.layer.set_contents(image.as_ptr() as id) };
+
+            transaction::commit();
+        }
     }
 
     pub(crate) unsafe fn set_buffer(&mut self, buffer: &[u32], width: u16, height: u16) {
-        let color_space = CGColorSpace::create_device_rgb();
-        let data =
-            unsafe { std::slice::from_raw_parts(buffer.as_ptr() as *const u8, buffer.len() * 4) }
-                .to_vec();
-        let data_provider = CGDataProvider::from_buffer(Arc::new(data));
-        let image = CGImage::new(
-            width as usize,
-            height as usize,
-            8,
-            32,
-            (width * 4) as usize,
-            &color_space,
-            kCGBitmapByteOrder32Little | kCGImageAlphaNoneSkipFirst,
-            &data_provider,
-            false,
-            kCGRenderingIntentDefault,
-        );
-
-        // The CALayer has a default action associated with a change in the layer contents, causing
-        // a quarter second fade transition to happen every time a new buffer is applied. This can
-        // be mitigated by wrapping the operation in a transaction and disabling all actions.
-        transaction::begin();
-        transaction::set_disable_actions(true);
-
-        unsafe { self.layer.set_contents(image.as_ptr() as id) };
-
-        transaction::commit();
+        self.resize(width.into(), height.into());
+        self.buffer_mut().copy_from_slice(buffer);
+        self.present();
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,6 +24,7 @@ mod error;
 mod util;
 
 use std::marker::PhantomData;
+use std::num::NonZeroU32;
 use std::ops;
 #[cfg(any(wayland_platform, x11_platform))]
 use std::rc::Rc;
@@ -76,7 +77,7 @@ macro_rules! make_dispatch {
         }
 
         impl SurfaceDispatch {
-            pub fn resize(&mut self, width: u32, height: u32) -> Result<(), SoftBufferError> {
+            pub fn resize(&mut self, width: NonZeroU32, height: NonZeroU32) -> Result<(), SoftBufferError> {
                 match self {
                     $(
                         $(#[$attr])*
@@ -297,7 +298,7 @@ impl Surface {
     /// in the upper-left corner of the window. It is recommended in most production use cases
     /// to have the buffer fill the entire window. Use your windowing library to find the size
     /// of the window.
-    pub fn resize(&mut self, width: u32, height: u32) -> Result<(), SoftBufferError> {
+    pub fn resize(&mut self, width: NonZeroU32, height: NonZeroU32) -> Result<(), SoftBufferError> {
         self.surface_impl.resize(width, height)
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,8 +20,10 @@ mod win32;
 mod x11;
 
 mod error;
+mod util;
 
 use std::marker::PhantomData;
+use std::ops;
 #[cfg(any(wayland_platform, x11_platform))]
 use std::rc::Rc;
 
@@ -44,7 +46,7 @@ macro_rules! make_dispatch {
     (
         $(
             $(#[$attr:meta])*
-            $name: ident ($context_inner: ty, $surface_inner : ty),
+            $name: ident ($context_inner: ty, $surface_inner: ty, $buffer_inner: ty),
         )*
     ) => {
         enum ContextDispatch {
@@ -82,16 +84,43 @@ macro_rules! make_dispatch {
                 }
             }
 
-            pub fn buffer_mut(&mut self) -> Result<&mut [u32], SoftBufferError> {
+            pub fn buffer_mut(&mut self) -> Result<BufferDispatch, SoftBufferError> {
                 match self {
                     $(
                         $(#[$attr])*
-                        Self::$name(inner) => inner.buffer_mut(),
+                        Self::$name(inner) => Ok(BufferDispatch::$name(inner.buffer_mut()?)),
+                    )*
+                }
+            }
+        }
+
+        enum BufferDispatch<'a> {
+            $(
+                $(#[$attr])*
+                $name($buffer_inner),
+            )*
+        }
+
+        impl<'a> BufferDispatch<'a> {
+            pub fn pixels(&self) -> &[u32] {
+                match self {
+                    $(
+                        $(#[$attr])*
+                        Self::$name(inner) => inner.pixels(),
                     )*
                 }
             }
 
-            pub fn present(&mut self) -> Result<(), SoftBufferError> {
+            pub fn pixels_mut(&mut self) -> &mut [u32] {
+                match self {
+                    $(
+                        $(#[$attr])*
+                        Self::$name(inner) => inner.pixels_mut(),
+                    )*
+                }
+            }
+
+            pub fn present(self) -> Result<(), SoftBufferError> {
                 match self {
                     $(
                         $(#[$attr])*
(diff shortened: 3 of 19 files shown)


ids1024:
Benchmarking is always subtle and easy to mess up, but comparing the time `set_buffer` takes against `resize`/`buffer_mut`/`present`, it seems like in the `winit` example this has about an order of magnitude decrease in the overhead softbuffer adds? (very roughly 1ms to 100µs, for instance, depending on window size).

It makes sense that this would be pretty efficient. In the `winit` example we shouldn't ever wait on buffer releases since we only draw when the server requests a draw, and Wayland is generally an asynchronous protocol, so while there may be a few context switches we shouldn't actually need to wait on the compositor for anything. So copying the buffer was the main overhead here.

ids1024:
Added code for Windows, using `BitBlt`. Looks like the performance is better there, but not by as much?

I guess it would be possible to write some portable automated tests `winit::Window::set_inner_size` to test at a particular size and test resizing. Maybe a timer to try redrawing periodically without a resize (or https://github.com/rust-windowing/winit/issues/2412 if that was implemented). And something to calculate average/min/max/etc. times for each function call and the total.

src/wayland/buffer.rs
```diff
@@ -150,6 +167,14 @@ impl WaylandBuffer {
...
         self.released.load(Ordering::SeqCst)
     }
+
+    fn len(&self) -> usize {
+        self.width as usize * self.height as usize
+    }
+
+    pub unsafe fn mapped_mut(&mut self) -> &mut [u32] {
+        unsafe { slice::from_raw_parts_mut(self.map.as_mut_ptr() as *mut u32, self.len()) }
+    }
```
Drakulix, line 177: This should be guarded by and error out, if the buffer was not previously released.

Note, that on some wayland compositors you don't even need a back-buffer, as the compositor will release any submitted memory buffer as soon as it has uploaded the buffer as a texture. Others will hold on to it and there is no guarantee (although all compositors do that in practice), that two buffers are enough.

In practice the client *should* keep a stack of buffers, creating new ones, if all are still hold by the compositor and re-using released ones instead, if possible.

ids1024: There's already code in `wayland/mod.rs` that waits for buffer release, so that should be handled already. Though there should be more comments about safety here.

> Note, that on some wayland compositors you don't even need a back-buffer

But there's no way the compositor explicitly communicates this, right? So we'll still end up allocating at least a second buffer if the application code wants to write to a buffer before it's been released. Which is likely if we can't guarantee that only happens on resize/frame callback/etc.

So as currently implemented this always uses two buffers for a surface, and blocks if the back buffer isn't released yet. I don't know if anything more complicated is typical or likely to be worthwhile.

Drakulix: > But there's no way the compositor explicitly communicates this, right?

Nope, but it also doesn't communicate, if it needs more than two. If you get a frame_callback, you should allocate a new buffer, if none are free. That is when you will know.
Technically blocking indefinitely is not against the spec, but allocating at max two buffers could render the app unresponsive on a valid compositor with an unusual implementation.

ids1024: Allowing allocating an unlimited number of buffers can cause an application submitting buffers too fast to crash. And it caused Gnome Shell to freeze temporarily: https://github.com/rust-windowing/softbuffer/issues/48

Looking at `mesa/src/egl/drivers/dri2/egl_dri2.h` and `mesa/src/egl/drivers/dri2/platform_wayland.c`, it seems it uses an array of 4 buffers? Then it blocks, though it uses a roundtrip:

```c
/* If we don't have a buffer, then block on the server to release one for
 * us, and try again. wl_display_dispatch_queue will process any pending
 * events, however not all servers flush on issuing a buffer release
 * event. So, we spam the server with roundtrips as they always cause a
 * client flush.
 */
if (wl_display_roundtrip_queue(dri2_dpy->wl_dpy,
                               dri2_surf->wl_queue) < 0)
    return -1;
```

notgull:
I added the X11 implementation for this. While we're rewriting the API, do we want to make these functions return `Result`s? On X11 at least there are a few failure states that can happen during the buffering process, and I either log or panic on them right now.

src/wayland/buffer.rs
```diff
@@ -150,6 +167,14 @@ impl WaylandBuffer {
     pub fn released(&self) -> bool {
         self.released.load(Ordering::SeqCst)
     }
+
+    fn len(&self) -> usize {
+        self.width as usize * self.height as usize
+    }
+
+    pub unsafe fn mapped_mut(&mut self) -> &mut [u32] {
+        unsafe { slice::from_raw_parts_mut(self.map.as_mut_ptr() as *mut u32, self.len()) }
```
i509VCB, line 176: MmapMut dereferences to `&mut [u8]`. Is there is particular reason we create the slice ourselves from the pointer?

src/lib.rs
```diff
@@ -235,6 +262,18 @@ impl Surface {
         })
     }
 
+    pub fn resize(&mut self, width: u32, height: u32) {
+        self.surface_impl.resize(width, height);
+    }
+
+    pub fn buffer_mut(&mut self) -> &mut [u32] {
```
i509VCB, line 269: What should occur if someone tries to acquire the buffer before the size was set.

src/lib.rs
```diff
@@ -235,6 +262,18 @@ impl Surface {
...
+        self.surface_impl.buffer_mut()
+    }
+
+    pub fn present(&mut self) {
+        self.surface_impl.present();
```
i509VCB, line 274: On Wayland this won't work unless the buffer size is at least 1x1, meaning a resize is needed before presenting. (You can present undefined contents however and be fine)./

src/wayland/buffer.rs
ids1024: We need an `&mut [u32]`, rather than `&mut [u8]`. So it needs an unsafe cast or transmute either way. We also want to restrict the size to the size of the buffer and not the total capacity of the pool.

So I don't think this is worse than other ways of doing the same thing, at least.

i509VCB: Ah okay, well this may change if we want more formats in the future.

src/lib.rs
ids1024: Yeah, I guess that isn't handled correctly at present. Though it's also an issue if `resize` is called with a size of `0`.

We could define the crate's API to initialize size to `(1, 1)`, and make `resize` take `NonZeroU32` ([Glutin seems to use this type](https://docs.rs/glutin/latest/glutin/surface/trait.GlSurface.html#tymethod.resize)). Or not have a default and require size arguments to `Surface::new`, but that's might be inconvenient if it's common to pass `1` and resize later.

Or we could allow size `0`, but just return a buffer of `&mut []`, and make present a no-op (may be unexpected if you've already set a buffer though; it could present a 1x1 buffer that's just white/alpha to at least clear the window). 

So this is a decision to make about the public API, and we can adjust the backends to match.

src/wayland/buffer.rs
ids1024: Yeah, if we want to support formats that aren't 32-bit per pixel, we'll need to figure out what API to expose for that.

notgull: Just a heads up: I already have `bytemuck` in the dependency tree, so you could probably use `bytemuck::cast_slice_mut` to avoid this unsafe.

src/wayland/buffer.rs
ids1024: Yeah, by default `EglSwapBuffers` blocks on a frame callback, unless you change `EglSwapInterval` to `0`. But that blocks indefinitely for windows that are occluded, so that's not great either.

The correct solution for the animation example is https://github.com/rust-windowing/winit/issues/2412.

ids1024:
> I added the X11 implementation for this. While we're rewriting the API, do we want to make these functions return `Result`s? On X11 at least there are a few failure states that can happen during the buffering process, and I either log or panic on them right now.

Yeah, the question is whether every method should return a `Result`, or if some are expected to never fail. Like `resize` is clearly infallible as implemented on Wayland now (just setting member variables). But that could fail if it actually performed resizing. Or maybe if the `u32` is out of range for `i32`, which Wayland uses (should never happen, so maybe better to just panic if anything).

For the public API, we should generally document what blocking, panicking, and errors are expected in each call. Including platform specific concerns if relevant.

notgull:
> Yeah, the question is whether every method should return a `Result`, or if some are expected to never fail.

At the very least, `present()` should return a `Result`, because that's usually where the most failure-prone API calls happen. I'm neutral on `resize()` and `buffer_mut()`. The way I have them implemented right now, errors *can* happen, it's just that unless there's an underlying I/O error or another call failed, it's unlikely to happen.

Also, I added support to the web backend for this PR.

src/wayland/buffer.rs
ids1024: Technically that adds overhead since `bytemuck` has to check alignment and panic if it's wrong. (And we know the memory is page aligned). Though the added cost is probably irrelevant.

This call is somewhat unsafe as implemented anyway though, since we're getting a mutable reference into shared memory, without checking that the buffer is released. Though I guess if we trust the server not to write to it anyway, any unsafely of mutating a buffer the server is reading isn't on our end...

src/win32.rs
```diff
@@ -7,28 +7,128 @@ use raw_window_handle::Win32WindowHandle;
...
+        unsafe {
+            Gdi::DeleteDC(self.dc);
+            Gdi::DeleteObject(self.bitmap);
+        }
+    }
+}
+
+impl Buffer {
+    fn new(window_dc: Gdi::HDC, width: i32, height: i32) -> Self {
+        let dc = unsafe { Gdi::CreateCompatibleDC(window_dc) };
```
i509VCB, line 42: Is this something we could expect to fail.

i509VCB: > ReleaseDC must be called from the same thread that called GetDC

That probably means !Send and !Sync

ids1024: It sounds like it's in principle more nuanced than that, and it's fine to access the DC from another thread, as long as it's only in use by one thread at a time and is released in the thread that created it?

But yeah, probably practically this means making `softbuffer::Surface` use `!Sync` and `!Send`. Do we expect that to be an issue for any uses? Should `softbuffer::Context` be usable in multiple threads? (So you could have a thread per window; but that does mean needing `Arc`/`Mutex` for shared state.)

Apparently `winit::event_loop::EventLoop` is `!Send` and `!Sync`. For similar platform reasons.

It looks like currently on Linux (on this branch) `Surface` is automatically `Send`/`Sync` if I compile only the Wayland backend, and `!Send`/`!Sync` if I compile with `x11`. So we probably want to be explicit and consistent about this.

ids1024:
So a lot of the questions here are tied to the design of the public API:

### Surface sizes
* Should we use `u32` or `NonZeroU32` for window sizes?
* `resize()` should possibly panic if the value is somehow too large. (Too large for an `i32` on Wayland; which would be an utterly absurd size. May be limited to 16-bit on some backends.)
* If a size of `0` is allowed, how does this behave? Including resizing from a larger size.
* Should `Surface::new` take `width`/`height`, or do we initialize to `0` or `1` and expect `resize` to be called later?

### Thread-safety
* `softbuffer::Surface` likely needs to be `!Send`/`!Sync` on some backends. This should be done explicitly and not backend-dependent.
* Is it necessary/desirable for `softbuffer::Context` to be `!Send`/`!Sync` too?

### Error handling
* `present()` should definitely return a `Result`. I guess `resize()` and `buffer_mut()` should be too, but whether they actually fail will depend on backend?
* Or `resize()` can be non-blocking and never fail, and defer any buffer allocating/resizing to the next `buffer_mut()` call.
* If we want to not panic even on things like disconnect from the display server, we need to return a `Result` or silently ignore errors in any call that might communicate with the display server. (A later call like `present` may still get an error.)

### Documentation
* We should document what errors are expected on functions returning `Result`. And what may cause panics, if applicable.
* The public API should document what assumptions are required for soundness. Like `buffer_mut()` trusts the display server not to mutate shared memory on backends using shared memory.
* What calls block? I guess we assume things like `buffer_mut()` won't block for long if the display server is well-behaved.

notgull:
My opinions, in no particular order:

> * `resize()` should possibly panic if the value is somehow too large. (Too large for an `i32` on Wayland; which would be an utterly absurd size. May be limited to 16-bit on some backends.)

We should return an error instead; the maximum size varies between backends, so this is a legitimate failure state rather than an unexpected issue. 

> * `present()` should definitely return a `Result`. I guess `resize()` and `buffer_mut()` should be too, but whether they actually fail will depend on backend?

`present()` and `resize()` should return a `Result`. `buffer_mut()` shouldn't and should just panic on failure. The case where is actually panics on failure (on X11 at least) is startingly rare.

> * Or `resize()` can be non-blocking and never fail, and defer any buffer allocating/resizing to the next `buffer_mut()` call.

I don't think this is a good idea; `buffer_mut()` should only block in the event that the SHM buffer is still being flushed out.

> * If we want to not panic even on things like disconnect from the display server, we need to return a `Result` or silently ignore errors in any call that might communicate with the display server. (A later call like `present` may still get an error.)

Silently ignoring errors is a recipe for disaster, I disagree with this on principle.

> * Should we use `u32` or `NonZeroU32` for window sizes?

`NonZeroU32` should be used, IMO.

> * `softbuffer::Surface` likely needs to be `!Send`/`!Sync` on some backends. This should be done explicitly and not backend-dependent.

Yes, we should use `PhantomData<*const ()>` or something to avoid surprises.

> * Is it necessary/desirable for `softbuffer::Context` to be `!Send`/`!Sync` too?

Yes, to future-proof it.

> * Should `Surface::new` take `width`/`height`, or do we initialize to `0` or `1` and expect `resize` to be called later?

`buffer_mut()` should panic if it's called before `resize()`.

ids1024:
> We should return an error instead; the maximum size varies between backends, so this is a legitimate failure state rather than an unexpected issue.

If the size is simply passed along from winit we know it must already be in range for the backend. (Are there expected uses where it might pass something out of range?). But returning an error is easy enough if we're making that function return a `Result` anyway. This would more be a consideration for if we didn't.

> buffer_mut() shouldn't and should just panic on failure. The case where is actually panics on failure (on X11 at least) is startingly rare.
> buffer_mut() should only block in the event that the SHM buffer is still being flushed out.

In the Wayland backend, `buffer_mut` currently also allocates and resizes the buffers. But with Wayland none of that should block (or I guess it would block of the socket write if the buffer was full, until the compositor reads?). And the only fallible part of that is allocating and mapping shared memory, which should only really happen when out of file descriptors / memory / virtual memory. So panics make sense.

That could be moved to `resize`. But regardless, both methods need to block on buffer release. Waiting for buffer release could block on the compositor for any amount of time but probably shouldn't block for long. And `wayland-client` could return a [`DispatchError`](https://docs.rs/wayland-client/latest/wayland_client/enum.DispatchError.html). So that needs to be turned into a `Err` return value, ignored, or turned into a panic.

> Silently ignoring errors is a recipe for disaster

Well, *specifically* if the error is permanent loss of connection to the display server, it would be fine since we'd be guaranteed to get the exact same error when `present` is called. But yeah, ignoring other errors would be problematic.

> Yes, to future-proof it.

It also lets use use `Rc`/`RefCell`/etc instead of `Arc`/`Mutex`.

> `buffer_mut()` should panic if it's called before `resize()`.

Hm. Hadn't thought of that option. I guess a default doesn't entirely make sense. Either that isn't what you wanted, and a panic would be clearer, or if you really want that you can just call `.resize(1, 1)`.

Currently in the `winit` example I have it calling `.resize()` and `.buffer_mut()` every draw. So another option would be to combine them into a single call, statically preventing this panic.

But if we add an option for `alpha` and even other (potentially not 32-bit) pixel formats, that also impacts the API here.

i509VCB:
We should probably have a test to ensure that the Send and Sync impls are correct.

I believe the static_assertions crate can do that

ids1024:
A simple macOS implementation for this API could just store a `Vec` that can be written into, then turned into a `CGDataProvider` and `CGImage`. So exactly like the current version but saving a copy. Swapping `IOSurface` front/back buffers may be viable as a more optimal solution (And these APIs seem to exist on iOS too).

@jackpot51 I'm not sure how this would be handled in the Orbital implementation. Unlike the others it fetches the current size of the window and maps it, then iterates over `y` copying each line. Rather than mapping a buffer for a exactly the `width`/`height` passed. (Also does this handle synchronization in some implicit way, or is there a possible race if the window is resized between getting the size and mapping, or while mapped?)

ids1024:
Oh, I guess the size of the window on orbital only changes when the client sends a call to change it.

So that makes sense but clashes somewhat with the way the API here works. Probably the size should match, but we don't want to panic when a size that doesn't match the window size is passed.

madsmtm:
> We should probably have a test to ensure that the Send and Sync impls are correct.
> 
> I believe the static_assertions crate can do that

You can't really do that, since you'd need to specify a negative bound. However, you _can_ do:
```rust
/// ```compile_fail
/// use softbuffer::Context;
/// fn needs_send<T: Send>() {}
/// needs_send::<Context>();
/// ```
/// ```compile_fail
/// use softbuffer::Surface;
/// fn needs_send<T: Send>() {}
/// needs_send::<Surface>();
/// ```
extern "C" {}
```

>     * Should `Surface::new` take `width`/`height`, or do we initialize to `0` or `1` and expect `resize` to be called later?

`Surface::new` already takes `HasRawWindowHandle`, can't it just figure out the initial height itself from that? (At least, I know it can on macOS).

> ### Thread-safety
> 
>     * `softbuffer::Surface` likely needs to be `!Send`/`!Sync` on some backends. This should be done explicitly and not backend-dependent.

While we're at it, `Surface::new` on macOS is only safe to call from the main thread, so perhaps we could enforce that as well? Though I guess I could also work around that by submitting work to the main thread queue instead.

i509VCB:
[assert_not_impl_all](https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_not_impl_all.html) is in static_assertions and should work.

madsmtm:
> [assert_not_impl_all](https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_not_impl_all.html) is in static_assertions and should work.

Huh, cool!

ids1024:
> While we're at it, Surface::new on macOS is only safe to call from the main thread, so perhaps we could enforce that as well? Though I guess I could also work around that by submitting work to the main thread queue instead.

Perhaps that should just be documented as a requirement of the unsafe `new()` functions. The caller is responsible for making sure the handles are valid as long as the `Context`/`Surface` exist. It make sense that would require they are called on the right thread if that is a requirement on the platform.

It would be nice if there could be an entirely safe API rather than the unsafely and unenforced lifetimes inherent in `raw-window-handle`... but I'm not sure how that could work across platforms.

ids1024:
Added an orbital implementation, though it still involves a copy. Now this should work on all backends. And CI is passing.

Is it useful to provide a copying API let `set_buffer` as well, or only the `buffer_mut()` API? It can perform better (copying once instead of twice) if a copy is still needed on certain backends, and the caller would just be copying from some other buffer into the one returned by `buffer_mut()`. And may be a slightly easier API to use in such cases.

But in most cases a client should be able to just render into `buffer_mut()`. I guess one exception would be if it wanted to persist the image data for the next draw.

notgull:
I think that we should just use `buffer_mut`. We may want to provide `set_buffer` for backwards compatibility, but if we do it should be marked as deprecated.

ids1024:
If it's just for backwards compatibility we can remove it. There aren't too many things using `softbuffer` currently, it's easy enough to adapt to, and we'll probably break other things before the next release anyway.

notgull:
What is the status of this PR? Is there anything I can do to advance it?

ids1024:
I'm still not sure about some of the details of `resize`/`buffer_mut`:
* On Wayland `buffer_mut()` is definitely fallible since it has to wait for buffer release, and could get a broken connection to the server, a protocol error, etc. So I guess it has to return a `Result` if we don't want these things to cause panics.
* Taking a `NonZeroU32` makes sense semantically but is a bit annoying when winit provides the size as a `u32`. And if it still returns an error when the size is too large for the backend.
* Would something like `buffer_mut(width: u32, height: u32) -> Result<&mut [u32], SoftBufferError>` be better than a separate `resize`? That would be suitable in the current examples at least, and avoids questions of dealing with not calling resize first.
    - But the API is going to become more complicated if other formats (at least an alpha option) are added anyway. So maybe better to keep them separate but make both return `Result` if needed.

I guess the copying in the orbital backend could be addressed by returning mapped memory if the size is the same and using the current copy approach only if it isn't.

ids1024:
Probably better to not worry to much and get this finished and merged. We can make more breaking API changes if needed later.

jackpot51:
@ids1024 on Orbital, the mapping the window buffer is safe only while the window resize is not acknowledged. Currently, winit does this whenever a resize event is read. Basically, the resize event tells the app what size the window manager requested, and then the app controls when to apply it. Any time the window manager is told to resize the window, any mappings become invalid.

I like having buffer_mut be provided the current width and height. This ensures that in most cases, it will not be possible for there to be a mismatch and the mapping can be valid. Although I'd prefer for the lifetime of the mapped buffer to be controlled more tightly, perhaps returning anything that can DerefMut to &mut [u32].

ids1024:
I also kind of like the option of having:

```rust
fn buffer_mut(&mut self, width: u32, height: u32)
    -> Result<Buffer<'_>, SoftBufferError>
```

And for `Buffer<'_>`:

```rust
fn present(self) -> Result<(), SoftBufferError>
```

A bit like `wgpu::SurfaceTexture::present()`. This makes it impossible to call `buffer_mut()` without setting the size, and impossible to call `present` without getting a buffer reference first.

Here `buffer_mut` has similar semantics to `wgpu::Surface::get_current_texture()` but with the addition of size arguments. Not sure if a different name would be better. (It shouldn't be called a "texture" anyway.)

We should also consider what variants to include in `SoftBufferError`  to make it practical for applications that want to actually handle errors in some way.

ids1024:
Implementing such a buffer type, with a `DerefMut` implementation, is a bit awkward since it's necessary to implement `Deref` too (taking `&self`), and the buffer type can't contain a borrow of the surface and of a part of the surface (for the buffer slice).

But I think it's ultimately a good API solution.

ids1024:
Perhaps it would be good to internally use a trait to implement backends. Something like this could work. While the public API would wrap and hide this.

```rust
trait Backend {
    type RawDisplay;
    type RawSurface; 
    type Display;
    type Surface;
    type Buffer;                                                                                  
    
    fn create_display(raw_display: Self::RawDisplay) -> Result<Self::Display, SoftBufferError>;
    fn create_surface(display: &Display, raw_surface: Self::RawSurface) -> Result<Self::Surface, SoftBufferError>;
    fn buffer_mut(surface: &mut Self::Surface, width: u32, height: u32) -> Result<Self::Buffer, SoftBufferError>;
    fn pixels(surface: &Self::Surface, buffer: &Self::Buffer) -> &[u32];
    fn pixels_mut(surface: &mut Self::Surface, buffer: &mut Self::Buffer) -> &mut [u32];
    fn present(surface: &mut Self::Surface, buffer: Self::Buffer) -> Result<(), SoftBufferError>;
}
```

ids1024:
The `Display` and `Buffer` associated types here could be `()` for backends that don't need any data associated with those. And this replaces the need to cover that in the invocation of `make_dispatch!`. Unfortunately there isn't any obvious way to avoid duplicating the list of backends in `Context::from_raw`/`Surface::from_raw`.

I guess if a `Buffer` type like this derefs, some caution is needed. If it does anything at all  non-trivial the overhead could be significant if the compiler can't eliminate the duplication when called for each pixel. (Even `.as_mut().unwrap()` on an option, but that would probably be optimized away in a loop?)

ids1024:
On second/third/fourth/fifth/... thought it's probably better to get this merged and resist my inclination to try to perfect every detail of the library's API in this PR.

Other things like support for other formats can be added in other PRs. At a minimum something simple to enable an alpha mode, though the API for buffers may be a bit different if it needed to support formats that aren't 32 bit-per-pixel.

madsmtm (approved):
macOS impl looks fine to me

src/x11.rs
```diff
@@ -187,138 +207,205 @@ impl X11Impl {
...
+            .or(Err(SoftBufferError::SizeOutOfRange { width, height }))?;
+        let height: u16 = height.try_into().or(Err(SoftBufferError::SizeOutOfRange {
+            width: width.into(),
+            height,
+        }))?;
+
+        if width != self.width || height != self.height {
+            self.buffer
+                .resize(&self.display.connection, width, height)
+                .map_err(|err| {
```
notgull, line 253: The `swbuf_err` method expresses this pattern more succinctly.

src/x11.rs
```diff
@@ -187,138 +207,205 @@ impl X11Impl {
...
-    ) -> Result<bool, PushBufferError> {
-        let shm_info = match self.shm {
-            Some(ref mut info) => info,
-            None => return Ok(false),
-        };
+    /// Get a mutable reference to the buffer.
+    pub(crate) fn buffer_mut(&mut self) -> Result<BufferImpl, SoftBufferError> {
+        log::trace!("buffer_mut: window={:X}", self.window);
+
+        Ok(BufferImpl(util::BorrowStack::new(self, |surface| {
```
notgull, line 272: I think that we can avoid this antipattern. If we call `finish_wait` before we return the buffer, it's guaranteed that Derefing the buffer will succeed.

notgull (changes requested):

src/x11.rs
ids1024: Really need some kind of benchmark to compare things like whether storing the slice helps.

The `BorrowStack` code should be safe, but admittedly is someone unusual and perhaps a bit confusing to read. Hopefully at some point Rust will have a better way to deal with this and self-referential structs.

README.md
```diff
@@ -72,21 +72,20 @@ fn main() {
...
                     (size.width, size.height)
                 };
-                let buffer = (0..((width * height) as usize))
-                    .map(|index| {
-                        let y = index / (width as usize);
-                        let x = index % (width as usize);
-                        let red = x % 255;
-                        let green = y % 255;
-                        let blue = (x * y) % 255;
+                surface.resize(width, height).unwrap();
```
john01dav, line 75: Do we want to resize on each render? Resizing on each resize, or perhaps checking for different sizing, is probably more reasonable. At present the cost of an extraneous resize is minimal, but that isn't documented as being the case so we should not rely on it. Documenting that the cost of a resize will be minimal is also an option, but this constricts future use so I advocate just fixing the examples (there are other examples that have this problem).

src/lib.rs
```diff
@@ -149,13 +199,15 @@ impl Context {
 
         Ok(Self {
             context_impl: imple,
+            _marker: PhantomData,
         })
     }
 }
 
 pub struct Surface {
```
john01dav, line 207: We need struct-level rustdoc

src/lib.rs
```diff
@@ -258,10 +319,25 @@ impl Surface {
...
     /// B: Blue channel
+    pub fn buffer_mut(&mut self) -> Result<Buffer, SoftBufferError> {
+        Ok(Buffer {
+            buffer_impl: self.surface_impl.buffer_mut()?,
+            _marker: PhantomData,
+        })
+    }
+}
+
+pub struct Buffer<'a> {
```
john01dav, line 330: We need struct-level rustdoc

src/lib.rs
```diff
@@ -272,15 +348,22 @@ impl Surface {
     ///
     /// If the caller wishes to synchronize other surface/window changes, such requests must be sent to the
     /// Wayland compositor before calling this function.
-    #[inline]
-    pub fn set_buffer(&mut self, buffer: &[u32], width: u16, height: u16) {
-        if (width as usize) * (height as usize) != buffer.len() {
-            panic!("The size of the passed buffer is not the correct size. Its length must be exactly width*height.");
-        }
+    pub fn present(self) -> Result<(), SoftBufferError> {
```
john01dav, line 351: It seems like it would be preferable to allow the user to create several buffers ahead of presenting them, to allow batching copies or (if we figure out a less restrictive threading model later) even preparing them on a background thread. For example, the animation example could create a series of Buffer objects, and then submit them reusably to be truly zero copy. To implement this we need present to take buffer as a reference, and we need to have it tied to a non-mut reference to surface. We also likely need some reference counting for the internal data to deal with keeping the buffers alive if the user drops them while the underlying API still needs them. With the Wayland backend at least, this seems possible, but please correct me if this is wrong.

src/lib.rs
```diff
@@ -272,15 +348,22 @@ impl Surface {
...
+    }
+}
 
-        unsafe {
-            self.surface_impl.set_buffer(buffer, width, height);
-        }
+impl<'a> ops::Deref for Buffer<'a> {
+    type Target = [u32];
+
+    fn deref(&self) -> &[u32] {
```
john01dav, line 359: Should be #[inline] to enable compiler optimizations.

john01dav (changes requested):
I mostly am commenting on the API, since I am not a platform expert for any of these platforms so I leave the comments on the implementations to each platform's maintainer. I would prefer if each platform's maintainer was asked to review these changes, however.

src/lib.rs
madsmtm: Plus a root-level `#![warn(missing_docs)]` so that it doesn't get missed again

src/cg.rs
```diff
@@ -35,22 +46,53 @@ impl CGImpl {
...
+        Ok(Self {
+            layer,
+            window,
+            color_space,
+            width: 0,
+            height: 0,
+        })
     }
 
-    pub(crate) unsafe fn set_buffer(&mut self, buffer: &[u32], width: u16, height: u16) {
```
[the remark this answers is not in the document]
ids1024: It's less efficient, on backends that can avoid a copy. So the question is how often it is likely to be used in practice.

src/lib.rs
ids1024: Something like that is definitely possible on Wayland. But on other platforms we can't call any/most window management related functions from off the main thread. So in a multi-threaded context even implementing `Drop` may be problematic if a `Buffer` can be used on another thread; maybe if it sets a flag and depends on one of these functions called in the main thread to eventually clean up resources.

If this is allowed only in a single thread I don't know if there are particular difficulties with it. Though it's non-trivial to support across backends.

src/x11.rs
ids1024: Done.

We'll want to think about how best to define the error type (or use multiple error types?) and move any helpers like this to `error.rs` and use them in all the backends. But that can be done later.

src/lib.rs
ids1024: In any case, this is a big change that is probably better left for a later pull request. It would probably be a breaking change, but at least until the next release we can break the API as many times as we want.

src/x11.rs
ids1024: Opened https://github.com/rust-windowing/softbuffer/issues/73.

src/cg.rs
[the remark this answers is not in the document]
ids1024: It would be trivial to implement `set_buffer` on top of the no-copy API, though that would be less efficient on backends that can't do no-copy presentation. Which seems bad, so it would be a bit of code that has to be maintained in every backend and in `lib.rs`.

I don't think the `fruit` example really represent a real-world, and I'm not sure the `animation` example really represents a good way to do animation in production use (though it didn't perform well when I tried to change it to use no-copy drawing without pre-rendering).

ids1024:
Did I really forget to add any doc comment at all to the `Surface` or `Buffer`? Oh well, documentation should be okay now but can always be improved.

Not sure what else we want to change before merging this. I expect we'll still want to make breaking changes (possibly pretty major ones) to the API before we settle on exactly how the crate should work. But at least once this is merged it will unblock work on some smaller things without conflicting with this.

notgull (approved):
I'll fix the weird stack borrowing in a follow-up PR.

ids1024:
Hey, but it's a fun weird borrow hack! And almost certainly™ sound. :laughing: 

Really need some kind of benchmark to empirically measure how that compares to alternatives, and likewise for other API design choices. On Linux deref will also have a branch for either Wayland or X... which probably has essentially zero impact but if it did end up testing that every pixel it might. The borrowing magic could be brought up to the top level of the library instead of in each backend....

Anyway, we can merge this PR without declaring the API perfect and final yet, and certainly without being entirely happy about implementation details. Currently merging is blocked by @john01dav's review. My inclination is that `set_buffer` is probably not useful or a good practice generally in real world use and doesn't need to be provided by the library.

notgull:
@john01dav By any chance, do you mind re-reviewing this? I have a [rendering crate](https://github.com/notgull/theo) that depends on the latest changes to this crate that I'd like to release.

ids1024:
There are some other things I'd like to work on in Softbuffer, but I wouldn't really want to work on any non-trivial changes until this is merged.

src/cg.rs
```diff
@@ -35,22 +46,55 @@ impl CGImpl {
...
-        let color_space = CGColorSpace::create_device_rgb();
-        let data =
-            unsafe { std::slice::from_raw_parts(buffer.as_ptr() as *const u8, buffer.len() * 4) }
-                .to_vec();
-        let data_provider = CGDataProvider::from_buffer(Arc::new(data));
+    pub fn resize(&mut self, width: u32, height: u32) -> Result<(), SoftBufferError> {
```
kchibisov, line 59: I'd suggest to use `NonZeroU16` and that way we could get rid of new error variant all together, though, the `Result` will still stay here. Or `NonZeroU32`.

src/wayland/mod.rs
```diff
@@ -15,7 +15,7 @@ struct State;
 
 pub struct WaylandDisplayImpl {
     conn: Connection,
-    event_queue: Mutex<EventQueue<State>>,
+    event_queue: RefCell<EventQueue<State>>,
```
kchibisov, line 18: You don't need this `RefCell`, so remove it.

src/wayland/mod.rs
```diff
@@ -72,60 +73,119 @@ impl WaylandImpl {
...
+    pub fn pixels(&self) -> &[u32] {
+        self.0.member()
+    }
+
+    #[inline]
+    pub fn pixels_mut(&mut self) -> &mut [u32] {
+        self.0.member_mut()
+    }
+
+    pub fn present(self) -> Result<(), SoftBufferError> {
```
kchibisov, line 150:
```suggestion
    pub fn present(mut self) -> Result<(), SoftBufferError> {
```
This should allow remove `RefCell`.

src/wayland/mod.rs
```diff
@@ -72,60 +73,119 @@ impl WaylandImpl {
...
+    pub fn buffer_mut(&mut self) -> Result<BufferImpl, SoftBufferError> {
+        let (width, height) = self
+            .size
+            .expect("Must set size of surface before calling `buffer_mut()`");
+
+        if let Some((_front, back)) = &mut self.buffers {
+            // Block if back buffer not released yet
             if !back.released() {
-                let mut event_queue = self.display.event_queue.lock().unwrap();
+                let mut event_queue = self.display.event_queue.borrow_mut();
```
kchibisov, line 100: This is behind `&mut`, so you can `get_mut()` directly.

kchibisov (changes requested):
Few notes about Wayland impl.

src/cg.rs
ids1024: There's an issue open about the fact the existing API only allows 16-bit values: https://github.com/rust-windowing/softbuffer/issues/13

I don't really expect anyone needs windows that large currently or in the foreseeable future, but it's probably best to stick with 32-bit if that's what modern backends support, and what winit, glutin, etc. use.

ids1024: Though it could be a panic instead of an error.

kchibisov: You can't even fit into memory such windows, they'll take 128GB for their frame buffer...

src/wayland/mod.rs
ids1024: As things are currently implemented the `EventQueue` is part of the `Rc<WaylandDisplayImpl>` shared by all surfaces using the display. And it needs a mutable references to dispatch. So `RefCell` is needed.

kchibisov: Aye.

src/wayland/mod.rs
kchibisov: You can still do that.

ids1024: It's not especially elegant, but it seems better than having softbuffer create an event queue per surface, and should work well enough in practice.

src/orbital.rs
```diff
@@ -42,50 +55,79 @@ impl Drop for OrbitalMap {
...
+        let (window_width, window_height) = self.window_size();
+
         {
             // Map window buffer
             let window_map =
-                unsafe { OrbitalMap::new(window_fd, window_width * window_height * 4) }
+                unsafe { OrbitalMap::new(self.window_fd(), window_width * window_height * 4) }
                     .expect("failed to map orbital window");
 
             // Window buffer is u32 color data in 0xAABBGGRR format
```
daxpedda, line 125: @jackpot51 sorry for my naivety here, but this comment suggests that the buffer taken and returned is partly reversed compared to what is expected by the rest of the library. Isn't that a problem?

src/lib.rs
```diff
@@ -232,38 +287,67 @@ impl Surface {
...
+        })
+    }
+}
+
+/// A buffer that can be written to by the CPU and presented to the window.
+///
+/// This derefs to a `[u32]`, which depending on the backend may be a mapping into shared memory
+/// accessible to the display server, so presentation doesn't require any (client-side) copying.
+///
+/// This trust the display server not to mutate the buffer, which could otherwise be unsound.
```
daxpedda, line 320:
```suggestion
/// This trusts the display server not to mutate the buffer, which could otherwise be unsound.
```

src/lib.rs
```diff
@@ -232,38 +287,67 @@ impl Surface {
...
+            buffer_impl: self.surface_impl.buffer_mut()?,
+            _marker: PhantomData,
+        })
+    }
+}
+
+/// A buffer that can be written to by the CPU and presented to the window.
+///
+/// This derefs to a `[u32]`, which depending on the backend may be a mapping into shared memory
+/// accessible to the display server, so presentation doesn't require any (client-side) copying.
```
daxpedda, line 318: I think it would be useful to add a `# Platform dependent behavior` documentation which platforms actually support no-copy presentation.

daxpedda (approved):
The web implementation seems LGTM to me.

---

Sorry to drop my 2c here a bit late to the party, but I was wondering why the original `Surface::set_buffer()` method was completely replaced instead of just amending `buffer_mut()` to `Surface`.

The issue is that if a user is unable to avoid not copying the buffer, for example they receive the buffer already as a `Vec` from somewhere else, if a platform doesn't support "no-copy" the user will be copying twice now: once from their buffer to the `softbuffer` buffer, then `softbuffer` will copy from it's buffer to the OS buffer.

This affects at least two Tier 1 platforms: X11 (sometimes) and MacOS. Additionally it affects Orbital (sometimes) and Web (in a weird way).

I propose to keep `Surface::set_buffer()` and maybe even go a step further, return an error in `buffer_mut()` if the surface can not be represented in a no-copy manner, making sure the user is aware and doesn't get the misconception that this is a no-copy operation.

---

This is off-topic for this PR, but originally I would have liked to introduce a new function in an extension trait for the Web target to be able to set the buffer in the correct format (RGBA vs 0RGB), avoiding the extra copying. This wouldn't play well with the missing `Surface::set_buffer()`.

I saw in https://github.com/rust-windowing/softbuffer/issues/17 that maybe there is the intention to introduce multiple `Surface` formats, which would solve that problem in a different, probably better way.

src/orbital.rs
ids1024: Hm, good point. In my testing it *seemed* like the colors where the same on all platforms, but something isn't right with the documentation here, at least.

ids1024: In any case the existing implementation isn't swapping components and this comment is already here, so this isn't something introduced in this PR. (Though if the format didn't match that might impact the possibility of no-copy on Orbital.)

ids1024:
Updated to use `NonZeroU32`. This would be more convenient if `winit` returned that instead of `u32`, but it's not too bad. We seem to be in agreement to leave out an API like `set_buffer()` for now, and change that later if it seems necessary.

kchibisov (changes requested):
Use `.get()` on `NonZero` types, instead of `u32::from`.

kchibisov (approved):

Issue #67, opened 2023-01-15
Handle scale factor change on macOS


kpreid:
The macOS code in softbuffer 0.2.0 consults `backingScaleFactor()` only when the graphics context is created. This means that, assuming the application always renders an image of the window's physical size, the image will display at the wrong scale (half or double) if the window has been dragged between monitors with different scale factors. Instead, the scale factor should probably be consulted at every `set_buffer()` call.

tronical:
Excellent idea to do that in `set_buffer`. I've made a PR that tries to do that.

PR #68, opened 2023-01-15
mac: Fix layer scale not being updated when contents is updated and window is on a new screen

tronical:
cc #67

### DIFF
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -16,6 +16,7 @@ use std::sync::Arc;
 
 pub struct CGImpl {
     layer: CALayer,
+    window: id,
 }
 
 impl CGImpl {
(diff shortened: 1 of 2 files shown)


tronical:
I hope that saving the window is ... safe enough in terms of memory management. Otherwise this seemed like an approach simple enough :)

src/cg.rs
```diff
@@ -27,14 +28,13 @@ impl CGImpl {
...
             layer.set_needs_display_on_bounds_change(false);
-            layer.set_contents_scale(window.backingScaleFactor());
             subview.setLayer(layer.id());
             subview.setAutoresizingMask_(NSViewWidthSizable | NSViewHeightSizable);
 
             view.addSubview_(subview); // retains subview (+1) = 2
             let _: () = msg_send![subview, release]; // releases subview (-1) = 1
         }
-        Ok(Self { layer })
+        Ok(Self { layer, window })
```
madsmtm, line 37: Wrt. memory management: `AppKitWindowHandle` should ideally guarantee that the window is alive, but I think I'd prefer a `let window: id = msg_send![window, retain];` before we store it, and a `let _: () = msg_send![self.window, release];` in `CGImpl`'s `Drop`.

madsmtm (changes requested):
Thanks!

src/cg.rs
tronical: Thanks, that's a good idea. Done.

madsmtm (changes requested):
Nice, could you add a changelog entry as well?

tronical:
Sure :)

madsmtm (approved):
Wonderful, thank you!

PR #69, opened 2023-01-23
fix: update width and height when resizing buffer

wash2:
This caused some issues for me when resizing a surface, which seem to be fixed by updating the width and height during a resize.


jackpot51 (approved):

ids1024:
:+1: I noticed this working on https://github.com/rust-windowing/softbuffer/pull/65 and have the same change there.

Hopefully *now* everything is correct on Wayland.

PR #71, opened 2023-02-20
Update dependencies

ids1024:
Using the latest `winit` version in examples fixes resizing behavior on Wayland, so the workaround for that in the `winit` example is also removed here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ x11rb = { version = "0.11.0", features = ["allow-unsafe-code", "dl-libxcb", "shm
 fastrand = { version = "1.8.0", optional = true }
 
 [target.'cfg(target_os = "windows")'.dependencies.windows-sys]
-version = "0.42.0"
+version = "0.45.0"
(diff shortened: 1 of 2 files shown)


ids1024:
> error: package `jpeg-decoder v0.3.0` cannot be built because it requires rustc 1.61.0 or newer, while the currently active rustc version is 1.60.0

Hm. I guess `image` can be left on the previous version.

notgull:
I feel like we can bump to 1.61.

ids1024:
Probably, though it also doesn't matter for a dev-depedency where the new version probably doesn't make any difference to the examples using it.

It seems good to keep `softbuffer` in sync with `winit`'s Rust MSRV. I guess both can be updated when there's some more significant reason to do so.

notgull:
`winit` will probably bump its MSRV to 1.61 soon anyways (rust-windowing/winit#2614 and rust-windowing/winit#2679), so I think a one-version bump wouldn't be that bad.

jackpot51 (approved):

PR #72, opened 2023-02-20
Add `run-wasm` workspace member, as used in `winit`

ids1024:
This makes it easy to run examples with the web backend, in the same way that winit examples are run.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+run-wasm = ["run", "--release", "--package", "run-wasm", "--"]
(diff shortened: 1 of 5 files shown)


john01dav:
Can we add a readme section with the exact command(s) to run this, including which rustup toolchains are needed? I found that `cargo run-wasm` and `cargo run-wasm --package run-wasm` did not work. I could spend a bunch more time struggling with this, but I am not likely to be the only one to be unsure how to do it and you presumably tested this and thus know so you can easily add it to the readme.

ids1024:
Added a quick mention to the README, like the winit README has.

notgull (approved):

PR #74, opened 2023-02-28
added `from_canvas` and `from_offscreen_canvas` methods to `Surface`

Toniman575:
Added methods to create `Surface` from a `HtmlCanvasElement` or `OffscreenCanvas` following wgpu's example https://github.com/gfx-rs/wgpu/blob/v0.15.1/wgpu/src/lib.rs#L1570-L1640

My use case required this to render in a web worker.

Edit: See this for a future alternative implementation https://github.com/rust-windowing/raw-window-handle/issues/102

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -235,6 +235,7 @@ impl Surface {
         })
     }
 
+    /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`](web_sys::HtmlCanvasElement).
     #[cfg(target_arch = "wasm32")]
     pub fn from_canvas(canvas: web_sys::HtmlCanvasElement) -> Result<Self, SoftBufferError> {
         let imple = SurfaceDispatch::Web(web::WebImpl::from_canvas(canvas)?);
@@ -244,6 +245,7 @@ impl Surface {
         })
     }
 
+    /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`](web_sys::OffscreenCanvas).
     #[cfg(target_arch = "wasm32")]
     pub fn from_offscreen_canvas(
         offscreen_canvas: web_sys::OffscreenCanvas,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -235,6 +235,26 @@ impl Surface {
         })
     }
 
+    #[cfg(target_arch = "wasm32")]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -235,6 +235,28 @@ impl Surface {
...
 
+    /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`](web_sys::HtmlCanvasElement).
+    #[cfg(target_arch = "wasm32")]
+    pub fn from_canvas(canvas: web_sys::HtmlCanvasElement) -> Result<Self, SoftBufferError> {
+        let imple = SurfaceDispatch::Web(web::WebImpl::from_canvas(canvas)?);
+
+        Ok(Self {
+            surface_impl: Box::new(imple),
+        })
+    }
```
notgull, line 246: The general pattern in `winit` is to make this an extension trait on the existing type. Could we do that here?

src/lib.rs
```diff
@@ -235,6 +235,28 @@ impl Surface {
...
+
+    /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`](web_sys::OffscreenCanvas).
+    #[cfg(target_arch = "wasm32")]
+    pub fn from_offscreen_canvas(
+        offscreen_canvas: web_sys::OffscreenCanvas,
```
notgull, line 251: I'd rather avoid exposing a public dependency on `web-sys` if at all possible (in case we decide to move to `stdweb` or something). Is there a way of representing these types without `web-sys`?

notgull (changes requested):

src/lib.rs
daxpedda: I don't know of a way to avoid that. Even [Winit does it](https://docs.rs/winit/0.28.1/wasm32-unknown-unknown/winit/platform/web/trait.WindowExtWebSys.html#tymethod.canvas).

As far as I know `stdweb` is dead, but if you know of an alternative to the `wasm-bindgen` ecosystem I would be very interested!

daxpedda: I forgot to mention, it is actually possible to represent these types without `web-sys`: using `JsValue`.
But this would still require `wasm-bindgen` as a public dependency, which doesn't really help if we want to change to an alternative later.

Toniman575:
Closing this PR in favor of #76

PR #76, opened 2023-02-28
Add `from_canvas()` to `Surface` for Wasm

daxpedda:
> Added methods to create `Surface` from a `HtmlCanvasElement` or `OffscreenCanvas` following wgpu's example https://github.com/gfx-rs/wgpu/blob/v0.15.1/wgpu/src/lib.rs#L1570-L1640
> 
> My use case required this to render in a web worker.
> 
> Edit: See this for a future alternative implementation [rust-windowing/raw-window-handle#102](https://github.com/rust-windowing/raw-window-handle/issues/102)

This PR addresses the issues in #74 and additionally handles an error when using `OffscreenCanvas`, which has a different context type: `OffscreenCanvasRenderingContext2d`. It was assumed that it inherits from `CanvasRenderingContext2d`, which it does not, see the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvasRenderingContext2D).

I spoke to @Toniman575 after we discussed some issues, he told me to replace his PR instead.

Replaces #74.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ wasm-bindgen = "0.2.78"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies.web-sys]
 version = "0.3.55"
-features = ["CanvasRenderingContext2d", "Document", "Element", "HtmlCanvasElement", "ImageData", "Window"]
+features = ["CanvasRenderingContext2d", "Document", "Element", "HtmlCanvasElement", "ImageData", "OffscreenCanvas", "Window"]
 
 [target.'cfg(target_os = "redox")'.dependencies]
 redox_syscall = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,4 +89,10 @@ members = [
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
-targets = ["x86_64-unknown-linux-gnu", "wasm32-unknown-unknown"]
+default-target = "x86_64-unknown-linux-gnu"
+targets = [
+    "x86_64-pc-windows-msvc",
+    "x86_64-apple-darwin",
+    "x86_64-unknown-linux-gnu",
+    "wasm32-unknown-unknown",
+]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ wasm-bindgen = "0.2.78"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies.web-sys]
 version = "0.3.55"
-features = ["CanvasRenderingContext2d", "Document", "Element", "HtmlCanvasElement", "ImageData", "OffscreenCanvas", "Window"]
+features = ["CanvasRenderingContext2d", "Document", "Element", "HtmlCanvasElement", "ImageData", "Window"]
 
 [target.'cfg(target_os = "redox")'.dependencies]
 redox_syscall = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,3 +85,8 @@ rayon = "1.5.1"
 members = [
     "run-wasm",
 ]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+targets = ["x86_64-unknown-linux-gnu", "wasm32-unknown-unknown"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,9 @@ use raw_window_handle::{
     HasRawDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle,
 };
 
+#[cfg(target_arch = "wasm32")]
+pub use self::web::SurfaceExtWeb;
+
 /// An instance of this struct contains the platform-specific data that must be managed in order to
 /// write to a window on that platform.
 pub struct Context {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 #![doc = include_str!("../README.md")]
 #![deny(unsafe_op_in_unsafe_fn)]
 #![warn(missing_docs)]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
 #[cfg(target_os = "macos")]
 #[macro_use]
--- a/src/web.rs
+++ b/src/web.rs
@@ -2,15 +2,20 @@
 
 #![allow(clippy::uninlined_format_args)]
 
+use js_sys::Object;
 use raw_window_handle::WebWindowHandle;
+use wasm_bindgen::prelude::wasm_bindgen;
 use wasm_bindgen::JsCast;
+use wasm_bindgen::JsValue;
 use web_sys::CanvasRenderingContext2d;
 use web_sys::HtmlCanvasElement;
 use web_sys::ImageData;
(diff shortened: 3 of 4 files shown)


src/lib.rs
```diff
@@ -30,6 +30,9 @@ use raw_window_handle::{
     HasRawDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle,
 };
 
+#[cfg(target_arch = "wasm32")]
+pub use self::web::SurfaceExtWeb;
```
daxpedda, line 34: Wasn't sure where to actually define and implement the trait, I hope this is alright.

src/web.rs
```diff
@@ -91,6 +147,39 @@ impl WebImpl {
             ImageData::new_with_u8_clamped_array(Clamped(&bitmap), width.into()).unwrap();
 
         // This can only throw an error if `data` is detached, which is impossible.
-        self.ctx.put_image_data(&image_data, 0.0, 0.0).unwrap();
+        match self {
+            WebImpl::Canvas { ctx, .. } => ctx.put_image_data(&image_data, 0.0, 0.0).unwrap(),
+            WebImpl::OffscreenCanvas { ctx, .. } => {
+                ctx.put_image_data(&image_data, 0.0, 0.0).unwrap();
+                ctx.commit();
```
daxpedda, line 154: Specifically for `OffscreenCanvas` I added calling `commit()` at the end of `set_buffer()` because otherwise nothing happens if `OffscreenCanvas` is used as a remote handle to an actual canvas, not sure if we want to leave that here or hide it behind another method.

src/web.rs
```diff
@@ -35,9 +39,32 @@ impl WebDisplayImpl {
...
+    fn put_image_data(
+        this: &OffscreenCanvasRenderingContext2d,
+        imagedata: &ImageData,
+        dx: f64,
+        dy: f64,
+    ) -> Result<(), JsValue>;
+
+    #[wasm_bindgen(method, structural, js_class = "OffscreenCanvasRenderingContext2D")]
+    pub fn commit(this: &OffscreenCanvasRenderingContext2d);
+}
```
daxpedda, line 57: `web-sys` doesn't currently support `OffscreenCanvasRenderingContext2D`, a PR has already been made and merged, but it will require a new version release: https://github.com/rustwasm/wasm-bindgen/pull/3319.

daxpedda:
CI failures are because `sudo apt-get update && sudo apt-get install gcc-multilib` fails on Linux.

ids1024:
CI seems to pass after rerunning the failed jobs.

daxpedda:
Thanks :tada:.

src/web.rs
daxpedda: Actually I just tried that users can do that on their own by just creating a separate context and calling `commit()` on that.

I found myself wanting to draw some text on the canvas with the browser API so I needed **not** to commit after calling `set_buffer()`.

src/lib.rs
notgull: This works for me!

src/web.rs
```diff
@@ -35,9 +39,29 @@ impl WebDisplayImpl {
     }
 }
 
-pub struct WebImpl {
-    canvas: HtmlCanvasElement,
-    ctx: CanvasRenderingContext2d,
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_name = OffscreenCanvasRenderingContext2D)]
+    pub type OffscreenCanvasRenderingContext2d;
```
madsmtm, line 45: This is was included in `web-sys` in https://github.com/rustwasm/wasm-bindgen/pull/3319, so I'd rather wait for that to be released than defining it ourselves.

madsmtm (changes requested):

src/web.rs
madsmtm: Let's wait for that, then

src/web.rs
```diff
@@ -91,6 +144,37 @@ impl WebImpl {
...
+    }
+}
+
+/// Extension methods for the Wasm target on [`Surface`](crate::Surface).
+pub trait SurfaceExtWeb: Sized {
+    /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`].
+    fn from_canvas(canvas: HtmlCanvasElement) -> Result<Self, SoftBufferError>;
+
+    /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`].
+    fn from_offscreen_canvas(offscreen_canvas: OffscreenCanvas) -> Result<Self, SoftBufferError>;
```
notgull, line 161: I'd still like to avoid having `web_sys` as a public dependency if possible. I know `winit` does it, but I'd like to keep the number of public dependencies in this project as low as possible.

Would it be possible to just pass in the ID/XPath of the canvas into these functions instead?

src/web.rs
daxpedda: Alright, will put it into draft mode then.

src/web.rs
madsmtm: That sounds reasonable to me

src/web.rs
daxpedda: So apparently `commit()` is deprecated, see https://github.com/whatwg/html/pull/3872.
It's not documented in the MDN, but apparently it works in both Firefox and Chrome, in Chrome `commit()` was actually removed entirely.

daxpedda:
Rebased after #77.

daxpedda:
Rebased after #88.

notgull:
Would it be possible to split this PR into one that adds the ability to use a `HtmlCanvasElement` and one for `OffscreenCanvas`? We can do the first one now but we need to wait until https://github.com/rustwasm/wasm-bindgen/pull/3319 for the second one.

Cargo.toml
```diff
@@ -85,3 +85,8 @@ rayon = "1.5.1"
 members = [
     "run-wasm",
 ]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+targets = ["x86_64-unknown-linux-gnu", "wasm32-unknown-unknown"]
```
daxpedda, line 92: The other targets don't have any target-specific APIs, should I still add them to the list?

daxpedda:
> Would it be possible to split this PR into one that adds the ability to use a `HtmlCanvasElement` and one for `OffscreenCanvas`? We can do the first one now but we need to wait until [rustwasm/wasm-bindgen#3319](https://github.com/rustwasm/wasm-bindgen/pull/3319) for the second one.

Done: https://github.com/rust-windowing/softbuffer/pull/103.

Cargo.toml
notgull: I would say yes; I think they've expected at this point.

src/web.rs
```diff
@@ -74,6 +72,21 @@ impl WebImpl {
         })
     }
 
+    fn resolve_ctx<T: JsCast>(
```
notgull, line 75: A comment explaining what this function does would be nice

notgull (approved):
LGTM aside from minor nitpicks

Cargo.toml
daxpedda: Only added 64-bit targets, say the word if I should add anything else.

src/web.rs
daxpedda: This function is useless right now, will re-introduce it in #103.
Removed it for now.

notgull (approved):

PR #77, opened 2023-02-28
Fix multi-threaded Wasm

daxpedda:
When I tried actually using Softbuffer with #76 in a multi-threaded Wasm environment using [`transferControlToOffscreen()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/transferControlToOffscreen) I found that Softbuffer actually fails on `ImageData::new_with_u8_clamped_array(wasm_bindgen::Clamped(&bitmap), width.into())`.

Apparently it is not allowed to create a `ImageData` from a `SharedArrayBuffer`. I only found a couple of Stackoverflow issues facing this ([#1](https://stackoverflow.com/questions/60911099), [#2](https://stackoverflow.com/questions/66827434)) but no actual documentation on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData) or [WHATWG](https://html.spec.whatwg.org/multipage/canvas.html#initialize-an-imagedata-object).

This fixes it by copying the data to a `Uint8Array` first and then creating the `ImageData` from that. Which is an extra copy that can't be avoided when using multi-threaded Wasm. I hid it all behind `#[cfg(target_feature = "atomics")]` to make sure not to affect people not needing this.

~~This can be further optimized by keeping the `Uint8Array` and not having to recreate it from scratch again every time. I thought maybe we should wait for #65 before that, happy to add it now if desired.~~
This could be further optimized if we add an additional function for Web that uses a copy API instead of the currently owned buffer one.

To make this happen I had to do an import by hand because `web-sys` doesn't actually provide the constructors we need for [`ImageData`](https://docs.rs/web-sys/0.3.61/web_sys/struct.ImageData.html#method.new_with_u8_clamped_array).

I also added `-Ctarget-feature=+atomics` to the CI separately to the regular Wasm build.

See the [`wasm-bindgen` documentation](https://rustwasm.github.io/docs/wasm-bindgen/examples/raytrace.html) for more information on multi-threaded Wasm.

This can be run locally with `RUSTFLAGS=-Ctarget-feature=+atomics,+bulk-memory cargo +nightly run-wasm --example winit -Zbuild-std=panic_abort,std`.

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -3,7 +3,6 @@
 #![allow(clippy::uninlined_format_args)]
 
 use raw_window_handle::WebWindowHandle;
-use wasm_bindgen::Clamped;
 use wasm_bindgen::JsCast;
 use web_sys::CanvasRenderingContext2d;
 use web_sys::HtmlCanvasElement;
@@ -142,9 +141,33 @@ impl<'a> BufferImpl<'a> {
             .flat_map(|pixel| [(pixel >> 16) as u8, (pixel >> 8) as u8, pixel as u8, 255])
             .collect();
 
+        #[cfg(target_feature = "atomics")]
+        let result = {
+            use js_sys::{Uint8Array, Uint8ClampedArray};
+            use wasm_bindgen::prelude::wasm_bindgen;
+            use wasm_bindgen::JsValue;
+
+            #[wasm_bindgen]
+            extern "C" {
+                #[wasm_bindgen(js_name = ImageData)]
+                type ImageDataExt;
+
+                #[wasm_bindgen(catch, constructor, js_class = ImageData)]
+                fn new(array: Uint8ClampedArray, sw: u32) -> Result<ImageDataExt, JsValue>;
+            }
+
+            let array = Uint8Array::new_with_length(bitmap.len() as u32);
+            array.copy_from(&bitmap);
+            let array = Uint8ClampedArray::new(&array);
+            ImageDataExt::new(array, self.imp.width)
+                .map(JsValue::from)
+                .map(ImageData::unchecked_from_js)
+        };
+        #[cfg(not(target_feature = "atomics"))]
+        let result =
+            ImageData::new_with_u8_clamped_array(wasm_bindgen::Clamped(&bitmap), self.imp.width);
         // This should only throw an error if the buffer we pass's size is incorrect.
-        let image_data =
-            ImageData::new_with_u8_clamped_array(Clamped(&bitmap), self.imp.width).unwrap();
+        let image_data = result.unwrap();
 
         // This can only throw an error if `data` is detached, which is impossible.
         self.imp.ctx.put_image_data(&image_data, 0.0, 0.0).unwrap();
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,11 +50,14 @@ jobs:
           # We're using Windows rather than Ubuntu to run the wasm tests because caching cargo-web
           # doesn't currently work on Linux.
           - { target: wasm32-unknown-unknown,   os: windows-latest,  }
+        include:
+          - rust_version: nightly
+          - { target: wasm32-unknown-unknown,   os: windows-latest, options: -Zbuild-std=panic_abort,std, rustflags: -Ctarget-feature=+atomics }
 
     env:
       RUST_BACKTRACE: 1
       CARGO_INCREMENTAL: 0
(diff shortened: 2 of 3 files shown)


daxpedda:
I am assuming that the CI test I added is not executed because of some security settings on the repo.

daxpedda:
Now that we decided that #76 should not be merged until `wasm-bindgen` releases a new version with the imports we needed (https://github.com/rust-windowing/softbuffer/pull/76#discussion_r1121792560), I rebased this PR onto master.

To address any potential concerns (considering @madsmtm didn't want to use custom imports but wait for `wasm-bindgen`/`web-sys`), the imports I introduced here won't make it into the `wasm-bindgen` ecosystem. This is because they already have their own custom variant of [`Uint8ClampedArray`](https://docs.rs/wasm-bindgen/0.2.84/wasm_bindgen/struct.Clamped.html) which really doesn't work the same way other `wasm-bindgen` imports work. This would require a breaking change on their side, not really something that will happen anytime soon.

src/web.rs
```diff
@@ -40,6 +40,26 @@ pub struct WebImpl {
...
+#[cfg(target_feature = "atomics")]
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_name = ImageData)]
+    type ImageDataExt;
+
+    #[wasm_bindgen(catch, constructor, js_class = "ImageData")]
+    fn new(array: Uint8ClampedArray, sw: u32) -> Result<ImageDataExt, wasm_bindgen::JsValue>;
+}
+
```
notgull, line 62: Is this not available in `wasm-bindgen` yet?

daxpedda: Looking at my own comment (https://github.com/rust-windowing/softbuffer/pull/77#issuecomment-1450402871) again, it actually didn't explain why this was necessary and I barely remembered anything myself. Apologies.

There is no `Uint8ClampedArray` in `web-sys`, there is [`Clamped`](https://docs.rs/wasm-bindgen/0.2.84/wasm_bindgen/struct.Clamped.html), which is supposed to fulfill the role of `Uint8ClampedArray`. I am assuming this was done for good reason: to implement the special behavior that is necessary to prevent copies when using it with slices.

The issue is that `Clamped` takes a generic, which can be a `Vec<u8>`, which presumably does what we want here, a copy, but `wasm-bindgen` doesn't support generics in their exports or imports. So, I presume, the decision was made to ignore the generic and only take `Clamped<&[u8]>`, which should be what you always want anyway.

Unfortunately, `ImageData` doesn't take a `Uint8ClampedArray` based on a `SharedArray`, which is what the underlying memory of `wasm-bindgen` is when you use the `atomics` target feature, and is what `Clamped` uses under the hood. This is why it will simply throw when using the `atomics` target feature. This is why a copy is necessary.

I checked again if this can somehow be upstreamed. The obvious solution is to implement generics, which is a huge project and probably doesn't have the reviewer capacity for it anyway.
The other option would be to adjust the WebIDL generator to include multiple methods to cover all these cases, but that would be very ugly and adds weird names, it's weird enough already, I will ask though.

I want to stress, though, that there is nothing wrong with adding custom `wasm-bindgen` definitions, it's the same mechanic `web-sys` uses too. It is frequently done by many projects to circumvent weird problems like this.

daxpedda: Ha, this is funny, I just started researching how complicated a fix would really be and found that `wasm-bindgen` itself is using the same workaround in their own example.

But (!), there's is simpler, because somehow I overlooked that `Uint8ClampedArray` is actually implemented, but in `js-sys`. Don't know how I overlooked this.

See https://github.com/rustwasm/wasm-bindgen/blob/e7c3f52e15f73702d8c94959574bcee1a0d86227/examples/raytrace-parallel/src/lib.rs#L119-L125.

I just pushed a fix for that.

daxpedda:
CI fails because the new version of `raw-window-handle` now requires Rust v1.64, but our CI runs on v1.60:
https://docs.rs/crate/raw-window-handle/0.5.1/source/Cargo.toml

notgull (approved):
If that's the blessed way of doing it, then I suppose everything else looks good enough.

notgull:
Try rebasing on top of the new `master`

daxpedda:
Done!

notgull (approved):

daxpedda:
Rebased after #65.

notgull:
This works on my machine under `run-wasm`, so I'll just merge it.

PR #82, opened 2023-03-20
feat: Use the safer API for window handles

notgull:
Builds on rust-windowing/raw-window-handle#110, rust-windowing/raw-window-handle#115 and rust-windowing/winit#2744 to see how the new safe raw window API does ergonomically.


notgull:
Ported to rust-windowing/raw-window-handle#116

notgull:
Superseded by #132

PR #85, opened 2023-04-06
Update `windows-sys` and `image`

ids1024:


notgull (approved):
Looks good to me!

Two ideas:

- We might want to use Dependabot to automate this process in the future.
- We might want to generate our own Windows bindings like `parking_lot` does to avoid `windows-rs` version churn, see microsoft/windows-rs#1720.

PR #86, opened 2023-04-06
m: Don't use the borrowing trick under X11

notgull:
Don't use the mapped handle trick under X11, since I'd like to avoid those kind of lifetime tricks if I can. Should hopefully make the code easier to reason about as a side effect.

It was proposed that this could cause a regression, so I also created a benchmark to test it. The results:

image

Not sure why this added a 3x speed decrease to `pixels_mut()`. It could be the `bytemuck` overhead, but I'd have to experiment to make sure.

### DIFF
--- a/benches/buffer_mut.rs
+++ b/benches/buffer_mut.rs
@@ -11,6 +11,7 @@ fn buffer_mut(c: &mut Criterion) {
     {
         use criterion::black_box;
         use softbuffer::{Context, Surface};
+        use std::num::NonZeroU32;
         use winit::platform::run_return::EventLoopExtRunReturn;
 
         let mut evl = winit::event_loop::EventLoop::new();
@@ -29,6 +30,14 @@ fn buffer_mut(c: &mut Criterion) {
                     let context = Context::new(elwt).unwrap();
                     Surface::new(&context, &window).unwrap()
                 };
+
+                let size = window.inner_size();
+                surface
+                    .resize(
+                        NonZeroU32::new(size.width).unwrap(),
+                        NonZeroU32::new(size.height).unwrap(),
+                    )
+                    .unwrap();
 
                 c.bench_function("buffer_mut()", |b| {
                     b.iter(|| {
--- a/benches/buffer_mut.rs
+++ b/benches/buffer_mut.rs
@@ -2,44 +2,52 @@ use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use softbuffer::{Context, Surface};
 
 fn buffer_mut(c: &mut Criterion) {
-    use winit::platform::run_return::EventLoopExtRunReturn;
+    #[cfg(any(target_arch = "wasm32", target_arch = "wasm64"))]
+    {
+        // Do nothing.
+    }
 
-    let mut evl = winit::event_loop::EventLoop::new();
-    let window = winit::window::WindowBuilder::new()
-        .with_visible(false)
-        .build(&evl)
-        .unwrap();
+    #[cfg(not(any(target_arch = "wasm32", target_arch = "wasm64")))]
+    {
+        use winit::platform::run_return::EventLoopExtRunReturn;
 
-    evl.run_return(move |ev, elwt, control_flow| {
-        control_flow.set_poll();
+        let mut evl = winit::event_loop::EventLoop::new();
+        let window = winit::window::WindowBuilder::new()
+            .with_visible(false)
+            .build(&evl)
+            .unwrap();
 
-        if let winit::event::Event::RedrawEventsCleared = ev {
-            control_flow.set_exit();
+        evl.run_return(move |ev, elwt, control_flow| {
+            control_flow.set_poll();
 
-            let mut surface = unsafe {
-                let context = Context::new(elwt).unwrap();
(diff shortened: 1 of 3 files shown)


ids1024:
You could try adding an `#[inline]` annotation to `buffer_mut` and `as_mut`, and see if that makes a difference.

Both of these functions have a `match`, so there's some branching overhead that the borrowing black magic avoided.

notgull:
> You could try adding an `#[inline]` annotation to `buffer_mut` and `as_mut`, and see if that makes a difference.
> 
> Both of these functions have a `match`, so there's some branching overhead that the borrowing black magic avoided.

It looks like adding inlining brings down the time to 357.63 ns, which I'd call "good enough". Not sure if it's the branch; when I replaced the `Wire` branch with `unreachable_unchecked()` it didn't have any kind of impact. It isn't `bytemuck` either, since I couldn't squeeze out any performance gains by removing it.

benches/buffer_mut.rs
```diff
@@ -0,0 +1,56 @@
...
+        evl.run_return(move |ev, elwt, control_flow| {
+            control_flow.set_poll();
+
+            if let winit::event::Event::RedrawEventsCleared = ev {
+                control_flow.set_exit();
+
+                let mut surface = unsafe {
+                    let context = Context::new(elwt).unwrap();
+                    Surface::new(&context, &window).unwrap()
+                };
```
ids1024, line 31: This panics when I run it on Wayland, since `resize` isn't being called before `buffer_mut`:

```
Benchmarking buffer_mut(): Warming up for 3.0000 sthread 'main' panicked at 'Must set size of surface before calling `buffer_mut()`', src/wayland/mod.rs:99:14
```

I guess for consistency we need to make sure that X (and any other backends) also panics here.

ids1024:
Comparing the current version of this branch with the previous `BorrowStack` version:

```
   Compiling softbuffer v0.2.0 (/home/ian/src/softbuffer)
    Finished bench [optimized] target(s) in 4.34s
     Running unittests src/lib.rs (target/release/deps/softbuffer-0bfdb4e10ca1ac68)

running 1 test
test util::tests::test_borrowstack_slice_int ... ignored

test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running benches/buffer_mut.rs (target/release/deps/buffer_mut-0d138639e721c38e)
buffer_mut()            time:   [5.7302 µs 5.7407 µs 5.7514 µs]
                        change: [-26.344% -26.183% -25.996%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low mild
  9 (9.00%) high mild

pixels_mut()            time:   [424.40 ns 424.53 ns 424.72 ns]
                        change: [+50.353% +50.733% +51.171%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 21 outliers among 100 measurements (21.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  18 (18.00%) high severe
```

Around 50% more time? So still a fair difference proportionally, though a lot better than 300%.

As for the absolute impact, if there was a deref for each pixel at 1080p, 424ns per pixel would come out to `(1080 * 1920 * 424) / 10**9` = .88 seconds? Which sounds unacceptably bad but also not representative of how softbuffer is actually performing, so we may need a different test to gauge how this works out in practice.

Where was that code I was working on to benchmark softbuffer... it seemed quite hard to actually get representative numbers and not see wildly different results for small variations in how the benchmarking is done.

notgull:
> As for the absolute impact, if there was a deref for each pixel at 1080p, 424ns per pixel would come out to `(1080 * 1920 * 424) / 10**9` = .88 seconds? Which sounds unacceptably bad but also not representative of how softbuffer is actually performing, so we may need a different test to gauge how this works out in practice.

This isn't actually the result of calling `buffer_mut()` once, but the benchmark calls `buffer_mut()` 500 times in a row. This way, I get a number that can be more accurately measured by the profiler.

https://github.com/rust-windowing/softbuffer/blob/a8b6491b0996e4c7531b16d77a2637beba1a9fa8/benches/buffer_mut.rs#L44-L47

Dividing by 500 gives us 1.76 ms... which is definitely not awful.

ids1024:
Ah, I didn't see that loop. That isn't so bad. Inverted that adds up to 568 frames per second. And 142 at 4k. If this accounted for most of the time per frame (unlikely), otherwise it's worse but this is likewise less significant.

Critereon's `Bencher::iter` is already meant to run things a bunch of times and get an average, right? So that alone isn't good enough here? Though really having a benchmark of any kind is a good start, and we can improve this in the future.

notgull:
> Critereon's `Bencher::iter` is already meant to run things a bunch of times and get an average, right? So that alone isn't good enough here? Though really having a benchmark of any kind is a good start, and we can improve this in the future.

If it only runs it once it only takes about 50 pictoseconds... at that point, you're less measuring the actual function and more so the overhead of `criterion` itself. If you're only running it once it can actually wildly vary between runs. Running it a bunch of times prevents this from happening.

ids1024 (approved):
Overall this seems okay. The impact shouldn't really be significant.

PR #87, opened 2023-04-06
Fix atomic Wasm CI

daxpedda:
I messed this up in #77.

On that note, where are we using `cargo web`, was this an old thing that got removed?
If yes, I can remove the comment and let it run on `ubuntu-latest` again. Probably remove the caching too.


notgull (approved):
Thanks!

PR #91, opened 2023-04-10
Add dependabot support to this repo

notgull:
Helps keep dependencies up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+    commit-message:
+      prefix: ''
+    labels: []
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
-    commit-message:
-      prefix: ''
-    labels: []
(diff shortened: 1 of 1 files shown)


john01dav (approved):

notgull:
> I don't know much about the details of dependabot, but this is a useful change. As such, I'm marking my review as positive, but if a review or feedback is wanted on the dependabot configuration it will need to come from someone else. I had it enabled on a previous repository (which seemed to happen automatically), and it sent me emails when it recommended updating. Who gets the emails in this case?

It opens a PR on this repository, so everyone who watches this repo (including me) gets an email.

ids1024:
Might as well do this. I haven't set up dependabot before on anything, but probably should.

ids1024:
Though actually: how exactly does dependabot work with Rust and semver? A library like this doesn't have a `Cargo.lock`. Does it try to update to the latest version and send a PR that may or may not still build successfully on CI?

notgull:
> Though actually: how exactly does dependabot work with Rust and semver? A library like this doesn't have a `Cargo.lock`. Does it try to update to the latest version and send a PR that may or may not still build successfully on CI?

You can see an example of this in #94. It updates to the latest version in the `Cargo.toml` in a PR. From there, CI tests it out.

ids1024:
Okay, so what I expected it to do. Which is useful but won't always work.

PR #95, opened 2023-04-10
WIP: Attempt to use `IOSurface` on macOS

ids1024:
It looks like `BytesPerRow` (the stride) has some kind of alignment requirement (testing on M1), that doesn't let us just use the width. So for this to work, softbuffer needs to expose the stride, and the user needs to deal with that...

This doesn't seem to be showing the right thing even when width matches stride. Not sure why.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,8 +47,10 @@ features = ["Win32_Graphics_Gdi", "Win32_UI_WindowsAndMessaging", "Win32_Foundat
 [target.'cfg(target_os = "macos")'.dependencies]
 bytemuck = { version = "1.12.3", features = ["extern_crate_alloc"] }
 cocoa = "0.24.0"
+core-foundation = "0.9.3"
 core-graphics = "0.22.3"
 foreign-types = "0.3.0"
+io-surface = "0.15.1"
 objc = "0.2.7"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
--- a/src/cg/buffer.rs
+++ b/src/cg/buffer.rs
@@ -1,0 +1,93 @@
+use core_foundation::{
+    base::TCFType, dictionary::CFDictionary, number::CFNumber, string::CFString,
+};
+use io_surface::{
+    kIOSurfaceBytesPerElement, kIOSurfaceBytesPerRow, kIOSurfaceHeight, kIOSurfacePixelFormat,
+    kIOSurfaceWidth, IOSurface, IOSurfaceRef,
+};
+use std::{ffi::c_int, slice};
+
+#[link(name = "IOSurface", kind = "framework")]
+extern "C" {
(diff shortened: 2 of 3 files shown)


ids1024:
> This doesn't seem to be showing the right thing even when width matches stride. Not sure why.

Ah, so this part is due to the alpha channel. https://developer.apple.com/documentation/corevideo/cvpixelformatdescription/1563591-pixel_format_identifiers doesn't list an `BGRX` format or similar, so it looks like we need to set the alpha of each pixel to `255` for it to display properly.

ids1024:
We'll probably need to expose a `stride` to efficiently implement https://github.com/rust-windowing/softbuffer/issues/44 using https://developer.android.com/ndk/reference/struct/a-native-window-buffer#struct_a_native_window___buffer

So I think we'll ultimately want an API change like this.

examples/winit.rs
```diff
@@ -42,14 +42,15 @@ fn main() {
...
                 let mut buffer = surface.buffer_mut().unwrap();
-                for index in 0..(width * height) {
-                    let y = index / width;
-                    let x = index % width;
-                    let red = x % 255;
-                    let green = y % 255;
-                    let blue = (x * y) % 255;
-
-                    buffer[index as usize] = blue | (green << 8) | (red << 16);
+                let stride = buffer.stride();
```
MarijnS95, line 45: Interesting, Android needs the same!

LoganDark:
> We'll probably need to expose a `stride` to efficiently implement #44 using https://developer.android.com/ndk/reference/struct/a-native-window-buffer#struct_a_native_window___buffer
> 
> So I think we'll ultimately want an API change like this.

I would highly recommend looking into [`imgref`](https://docs.rs/imgref/latest/imgref) for this.

ids1024:
Is `imgref` currently used in any crates that would likely be used with Softbuffer for rendering?

We need the `Buffer` type to provide a `present` method, and on some backends to release certain resources on drop. So it's natural enough to provide a `stride()` method on that. But we could add a method returning `ImgRefMut` if that will be useful to users of softbuffer.

ids1024:
> Bonus if you do ImgRefMut<ARGB> instead of ImgRefMut<u32> but either way is fine.

Yeah, that part gets back to https://github.com/rust-windowing/softbuffer/issues/98. Unfortunately it seems we need to use RGBA on some platforms for no-copy presentation....

So I think we've settled on this solution for macOS, it's just that it requires breaking API changes, and we may want to do that along with https://github.com/rust-windowing/softbuffer/issues/98.

LoganDark:
> Unfortunately it seems we need to use RGBA on some platforms for no-copy presentation

Well right now you say `u32`. Using the redicated structs from [`rgb`](https://docs.rs/rgb/latest/rgb/) crate would be better and it's even what I use internally before converting to `u32`. So in a way you just proven my point

madsmtm:
Superseded by https://github.com/rust-windowing/softbuffer/pull/329.

PR #96, opened 2023-04-11
WIP: macos: Copy to IOSurface on present

ids1024:
Alternative to https://github.com/rust-windowing/softbuffer/pull/95, that avoids API changes.

In a release build on an M1 mac, fullscreened at 1440p, this brings the `animation` examples redraw time from about 10ms to 900μs. But the `winit` example ends up increasing to 20ms from around 18ms. This does include resizing time, but seems to still take at least a ms more even excluding that?

https://github.com/rust-windowing/softbuffer/pull/95 does improve the `winit` example here to 12ms, but gets 1.25ms for `animation` (that's probably since it currently allocates an `IOBuffer` every single present, so it could be fixed by doing that correctly).

Interesting how this version compares. There is also likely some overhead to the previous version this test isn't catching (When does it actually copy from the `CGDataProvider` to GPU? Is that later in our main loop, or does the display server handle that?).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,6 @@ bytemuck = { version = "1.12.3", features = ["extern_crate_alloc"] }
 cocoa = "0.24.0"
 core-foundation = "0.9.3"
 core-graphics = "0.22.3"
-foreign-types = "0.3.0"
 io-surface = "0.15.1"
 objc = "0.2.7"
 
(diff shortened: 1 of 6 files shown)


ids1024:
Interesting, the main time overhead on the `animation` example with `master` that is fixed is in `transaction::commit`. Not sure what that is doing. It's not taking any significant amount of time with the `winit` example, so it presumably has something to do with drawing unthrottled when a redraw wasn't requested by the display server...

madsmtm:
I've added `byte_stride` in #315, so we're not going to go this route.

See https://github.com/rust-windowing/softbuffer/pull/329 for adding `IOSurface`.

PR #97, opened 2023-04-11
examples: Avoid dividing index to get x, y; calculate index from x, y

ids1024:
Division is more expensive than multiplication, and this doesn't seem to be optimize away, so iterating over `(0..height)` and `(0..width)` provides better performance.

It's good to have the example show the best way to write code using softbuffer, and this helps when using examples as a benchmark. This will also be needed if Softbuffer supports formats where `stride` isn't just `width * pixel_size`, and needs alignment.


notgull (approved):
LGTM

jackpot51 (approved):

Issue #39, opened 2022-12-21
Damage in presentation


i509VCB:
In order to efficiently draw on the CPU, it may be desireable to reduce how many pixels are actually drawn.

Often this involves telling the display server what pixels were changed so that the display server can reuse it's existing framebuffer or perform a partial upload of the texture data to the gpu to save pci bandwidth.

ids1024:
Now that https://github.com/rust-windowing/softbuffer/pull/65 is merged, we could have a `present_with_damage()` method.

This is natural enough on Wayland. Do other backends have a way to support this?

PR #99, opened 2023-04-21
Add `Buffer::present_with_damage()` and `Buffer::age()`

ids1024:
https://github.com/rust-windowing/softbuffer/issues/39.

Implemented only for Wayland. Should be able to add an X implementation using `XDamage`, but I haven't looked at how that works yet. On `web` we could use this to update only the changed part of the canvas possibly. If we need to convert the format anyway. I don't see any API for damage that could be used on macOS or Windows.

Also should be tested in some way.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -124,6 +124,15 @@ macro_rules! make_dispatch {
                 }
             }
 
+            pub fn age(&self) -> u8 {
+                match self {
+                    $(
+                        $(#[$attr])*
+                        Self::$name(inner) => inner.age(),
+                    )*
+                }
+            }
+
             pub fn present(self) -> Result<(), SoftBufferError> {
                 match self {
                     $(
@@ -326,7 +335,7 @@ impl Surface {
 
     /// Return a [`Buffer`] that the next frame should be rendered into. The size must
     /// be set with [`Surface::resize`] first. The initial contents of the buffer may be zeroed, or
-    /// may contain a previous frame.
+    /// may contain a previous frame. Call [`Buffer::age`] to determine this.
     pub fn buffer_mut(&mut self) -> Result<Buffer, SoftBufferError> {
         Ok(Buffer {
             buffer_impl: self.surface_impl.buffer_mut()?,
@@ -377,6 +386,16 @@ pub struct Buffer<'a> {
 }
 
 impl<'a> Buffer<'a> {
+    /// Is age is the number of frames ago this buffer was last presented. So if the value is
+    /// `1`, it is the same as the last frame, and if it is `2`, it is the same as the frame
+    /// before that (for backends using double buffering). If the value is `0`, it is a new
+    /// buffer that has unspecified contents.
+    ///
+    /// This can be used to update only a portion of the buffer.
+    pub fn age(&self) -> u8 {
+        self.buffer_impl.age()
+    }
+
     /// Presents buffer to the window.
     ///
     /// # Platform dependent behavior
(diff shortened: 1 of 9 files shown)


ids1024:
https://www.x.org/releases/X11R7.7/doc/damageproto/damageproto.txt
> Damage is automatically computed by the X Server for X rendering operations,
but direct rendering extensions have allowed clients to perform rendering
outside of the control of the X Server.  The 1.1 version of the protocol
added a request to allow direct rendering clients to report damage to a
drawable.

If we're using `put_image`/`shm_put_image`, we're not doing "direct rendering" (like a DRI based OpenGL client), and those calls already report damage to region we "put" to? And I don't see a way to disable that damage reporting.

So I guess on X, we'd not use XDamage here, but could call `put_image`/`shm_put_image` once per damage rectangle, with different `src_x`/`src_y`/`dest_x`/`dest_y`/`width`/`height` arguments?

ids1024:
Everything here *should* work now, but is untested.

src/x11.rs
```diff
@@ -294,13 +304,21 @@ impl<'a> BufferImpl<'a> {
         unsafe { self.0.buffer.buffer_mut() }
     }
 
+    pub fn age(&self) -> Option<NonZeroU8> {
```
kchibisov, line 307: I'd suggest to use `0` actually, since it'll naturally work. Like we have 0 in buffer_age ext.

ids1024: Yeah, after thinking about it, that is probably simpler.

src/lib.rs
```diff
@@ -207,6 +225,19 @@ impl Context {
...
+#[derive(Clone, Copy, Debug)]
+pub struct Rect {
+    /// x coordinate of top left corner
+    pub x: i32,
+    /// y coordinate of top left corner
+    pub y: i32,
+    /// width
+    pub width: i32,
+    /// height
+    pub height: i32,
```
daxpedda, line 238:
```suggestion
    /// x coordinate of top left corner
    pub x: u32,
    /// y coordinate of top left corner
    pub y: u32,
    /// width
    pub width: NonZeroU32,
    /// height
    pub height: NonZeroU32,
```
Wouldn't that make more sense?
I assume you decided on `i32` because all APIs take a `i32`, but Web doesn't. In any case, I think this would be better, I'm not actually sure what a negative `x`, `y`, `width` or `height` would do.

ids1024: Windows and Wayland take `i32` here. But I expect that's just C programmers defaulting to `int` even for things that should never be negative. Presumably none of these should have negative values, and width and height should be non-zero.

Either way it requires conversion depending on the backend (like the current `NonZeroU32` sizes). I guess it can return an error for values that are out of range...

ids1024:
Updated to use `u32`/`NonZeroU32` and return an error for out-of-range damage rects. It also now panics on present if `resize` wasn't called on X11 and Windows now, though we may want to require an initial window size instead (https://github.com/rust-windowing/softbuffer/issues/106).

Still not properly tested, but there *shouldn't* be much that could go wrong here, and merging would help us work on other things without accumulating more conflicts. I still need to see about making use of both of these features in Iced's new software renderer.

ids1024:
Updated to address conflicts with https://github.com/rust-windowing/softbuffer/pull/104.

notgull (approved):
Looks good to me!

ids1024:
I think we can merge this now. So more backend code changes can be made without adding conflicts. If any issues come up later, they can be fixed.

Issue #98, opened 2023-04-11
Pixel format API design


ids1024:
We need an API to list available pixel formats, and select one. Some considerations:
* We want a way to support transparency: https://github.com/rust-windowing/softbuffer/issues/17
* The web backend current converts BGRX to RGBA. It would be better if it could use RGBA directly, if an application can render that way.
* It looks like for https://github.com/rust-windowing/softbuffer/issues/44 we'd have to use RGBX, rather than BGRX.
* For https://github.com/rust-windowing/softbuffer/pull/95, it seems BGRA is supported, but not BGRX. So we need to make sure to set alpha to 255.
* Wayland's `wl_shm` can support many formats, but only two are required to be supported by compositors.

So some questions here:
* What formats should be supported?
   - We want to offer a format for each platform that can be used efficiently without conversion, at least, and there doesn't seem to be one that can be assumed everywhere, unfortunately.
   - There are *many* formats that could be supported on at least some platforms.
* Do we want to support formats that are not 4 byte per pixel? This will impact the API, which currently uses u32 values for pixels.
* How to we communicate what formats are supported, and what formats are optimal?
* We still want at least one format to work on all platforms, even if it requires conversion on some, right? What formats should be universally supported, then?

notgull:
[Here](https://github.com/bread-graphics/genimage/blob/master/src/format.rs) is an API I wrote a while back for an image crate I was writing (abandoned as of now). The idea is that is specifies RGBA channels, their sizes, their order and whether or not they are present. It's based on the setup that `pixman` uses. I think a similar type would be good for `softbuffer`.

What if we made it so, no matter what format the buffer is, it's transcoded to the proper format? However, the API would expose a "transcode-free formats" function (maybe an `Iterator`?) that indicates which formats aren't transcoded at all?

ids1024:
Technically that doesn't cover every possible format. Wayland for instance [supports](https://wayland.app/protocols/wayland#wl_shm:enum:format) things like YUV, indexed color, etc. But I doubt YUV is useful for softbuffer consumers, and indexed color (hopefully) isn't relevant these days, nor is it useful without knowing the palette. :laughing:

I guess that iterator would be returned by a method of `Surface`? And there would be a `set_format` method to set the format for the next `buffer_mut`? Unlike `resize` this call would be optional, and default to the current default.

But still, we have the fact `Buffer` derefs to `&[u32]`, which is a problem if we want to support formats with more or less than 32 bits per pixel.

[...]

PR #101, opened 2023-04-27
Implement `present_with_damage` for Web

daxpedda:
Currently everything is still converted to RGBA, even if we know which parts to convert. We can figure that out when we address #98.


ids1024:
Ah, I was thinking about converting only the pixels in the damage regions, but I didn't see there was an API that could be used to specify a damage region like this.

PR #103, opened 2023-05-08
Add `OffscreenCanvas` support

daxpedda:
This is blocked on ~~https://github.com/rustwasm/wasm-bindgen/pull/3319~~ https://github.com/rustwasm/wasm-bindgen/pull/3416. Waiting for a new `web-sys` release.

Based on #76 (merged).

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -2,11 +2,12 @@
 
 #![allow(clippy::uninlined_format_args)]
 
+use js_sys::Object;
 use raw_window_handle::WebWindowHandle;
-use wasm_bindgen::JsCast;
-use web_sys::CanvasRenderingContext2d;
-use web_sys::HtmlCanvasElement;
+use wasm_bindgen::{JsCast, JsValue};
 use web_sys::ImageData;
+use web_sys::{CanvasRenderingContext2d, HtmlCanvasElement};
+use web_sys::{OffscreenCanvas, OffscreenCanvasRenderingContext2d};
 
 use crate::error::SwResultExt;
 use crate::{Rect, SoftBufferError};
@@ -33,11 +34,8 @@ impl WebDisplayImpl {
 }
 
 pub struct WebImpl {
-    /// The handle to the canvas that we're drawing to.
-    canvas: HtmlCanvasElement,
-
-    /// The 2D rendering context for the canvas.
-    ctx: CanvasRenderingContext2d,
+    /// The handle and context to the canvas that we're drawing to.
+    canvas: Canvas,
 
     /// The buffer that we're drawing to.
     buffer: Vec<u32>,
@@ -47,6 +45,19 @@ pub struct WebImpl {
 
     /// The current canvas width/height.
     size: Option<(NonZeroU32, NonZeroU32)>,
+}
+
+/// Holding canvas and context for [`HtmlCanvasElement`] or [`OffscreenCanvas`],
+/// since they have different types.
+enum Canvas {
+    Canvas {
+        canvas: HtmlCanvasElement,
+        ctx: CanvasRenderingContext2d,
+    },
+    OffscreenCanvas {
+        canvas: OffscreenCanvas,
+        ctx: OffscreenCanvasRenderingContext2d,
+    },
 }
 
 impl WebImpl {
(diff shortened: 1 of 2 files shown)


src/web.rs
```diff
@@ -46,6 +47,18 @@ pub struct WebImpl {
     width: u32,
 }
 
+pub enum Canvas {
```
notgull, line 50: Would be nice to document this enum. Also, does it need to be `pub`?

src/web.rs
```diff
@@ -2,11 +2,15 @@
 
 #![allow(clippy::uninlined_format_args)]
 
+use js_sys::Object;
 use raw_window_handle::WebWindowHandle;
 use wasm_bindgen::JsCast;
+use wasm_bindgen::JsValue;
```
notgull, line 8: Could you merge these two includes into one statement? Makes it cleaner.

src/web.rs
```diff
@@ -105,6 +139,10 @@ impl WebImpl {
 pub trait SurfaceExtWeb: Sized {
     /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`].
     fn from_canvas(canvas: HtmlCanvasElement) -> Result<Self, SoftBufferError>;
+
+    /// Creates a new instance of this struct, using the provided [`HtmlCanvasElement`].
+    #[cfg(web_sys_unstable_apis)]
```
notgull, line 144: Is this `cfg` stable? If so, the method documentation should mention it (and it should also be `doc(cfg(..))`)

src/web.rs
```diff
@@ -46,6 +47,18 @@ pub struct WebImpl {
     width: u32,
 }
 
+pub enum Canvas {
+    Canvas {
+        canvas: HtmlCanvasElement,
+        ctx: CanvasRenderingContext2d,
+    },
+    #[cfg(web_sys_unstable_apis)]
```
daxpedda, line 55: Not exactly ideal, but it works and doesn't really affect the public API if a user doesn't use `cfg(web_sys_unstable_apis)`.
As ugly as this is, it's how `wasm-bindgen` works.

We can remove this in the next `web-sys` release (https://github.com/rustwasm/wasm-bindgen/pull/3416) or wait another turn for it, WDYT @notgull?

notgull (changes requested):

src/web.rs
notgull: I'd say to wait for it, since the next release shouldn't be that far away.

daxpedda: Alright then.

src/web.rs
daxpedda: It's Rust stable, but it's not stable API in `web-sys`.
> (and it should also be `doc(cfg(..))`)

Just tried it, `feature(doc_auto_cfg)` already takes care of that.

src/web.rs
daxpedda: Done.

I guess we don't want to switch to nightly Rustfmt for [`[imports_granularity](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_granularity)`](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_granularity)?

src/web.rs
daxpedda: Done.

src/web.rs
```diff
@@ -2,11 +2,14 @@
...
 
+use js_sys::Object;
 use raw_window_handle::WebWindowHandle;
-use wasm_bindgen::JsCast;
+use wasm_bindgen::{JsCast, JsValue};
 use web_sys::CanvasRenderingContext2d;
 use web_sys::HtmlCanvasElement;
 use web_sys::ImageData;
+use web_sys::OffscreenCanvas;
+use web_sys::OffscreenCanvasRenderingContext2d;
```
notgull, line 12: These should be combined into one block.

notgull (changes requested):

src/web.rs
daxpedda: Done.

notgull (approved):

daxpedda:
Let's merge it after https://github.com/rust-windowing/softbuffer/pull/104, rebasing #103 is easier.

notgull (approved):
Looks good to me!

PR #104, opened 2023-05-13
Add a function for retrieving the window contents

### LINKED ISSUE #36: Add a method for getting pixels from a window
It would be nice if there were a method for getting pixels from a window in addition to setting them. If you wanted to draw a transparent or blended image using software rendering, you would need access to the image currently in the window.

notgull:
Closes #36 by adding a `fetch()` method for the `Buffer` method. Might need a rework in the face of #98 or #99.

WIP for now, with only implementations for X11 and Win32. I would appreciate guidance from the platform maintainers on how this would be implemented on other platforms.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,6 +136,15 @@ macro_rules! make_dispatch {
                     )*
                 }
             }
+
+            pub fn fetch(&mut self) -> Result<(), SoftBufferError> {
+                match self {
+                    $(
+                        $(#[$attr])*
+                        Self::$name(inner) => inner.fetch(),
+                    )*
+                }
+            }
         }
     };
 }
@@ -373,6 +382,11 @@ impl<'a> Buffer<'a> {
     /// Wayland compositor before calling this function.
     pub fn present(self) -> Result<(), SoftBufferError> {
         self.buffer_impl.present()
+    }
+
+    /// Copies the window contents into this buffer.
+    pub fn fetch(&mut self) -> Result<(), SoftBufferError> {
+        self.buffer_impl.fetch()
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -321,6 +321,8 @@ impl Surface {
     ///
     /// - On X11, the window must be visible.
     /// - On macOS, Redox and Wayland, this function is unimplemented.
+    /// - On Web, this will fail if the content was supplied by
+    ///   a different origin depending on the sites CORS rules.
     pub fn fetch(&mut self) -> Result<Vec<u32>, SoftBufferError> {
         self.surface_impl.fetch()
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -320,6 +320,7 @@ impl Surface {
     /// ## Platform Dependent Behavior
     ///
     /// - On X11, the window must be visible.
+    /// - On macOS, Redox and Wayland, this function is unimplemented.
     pub fn fetch(&mut self) -> Result<Vec<u32>, SoftBufferError> {
         self.surface_impl.fetch()
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,6 +98,15 @@ macro_rules! make_dispatch {
                     )*
                 }
             }
+
+            pub fn fetch(&mut self) -> Result<Vec<u32>, SoftBufferError> {
+                match self {
+                    $(
+                        $(#[$attr])*
+                        Self::$name(inner) => inner.fetch(),
+                    )*
+                }
+            }
         }
 
         enum BufferDispatch<'a> {
@@ -133,15 +142,6 @@ macro_rules! make_dispatch {
                     $(
                         $(#[$attr])*
                         Self::$name(inner) => inner.present(),
-                    )*
-                }
-            }
-
-            pub fn fetch(&mut self) -> Result<(), SoftBufferError> {
-                match self {
-                    $(
-                        $(#[$attr])*
-                        Self::$name(inner) => inner.fetch(),
                     )*
                 }
             }
@@ -315,6 +315,15 @@ impl Surface {
         self.surface_impl.resize(width, height)
     }
 
+    /// Copies the window contents into a buffer.
+    ///
+    /// ## Platform Dependent Behavior
+    ///
+    /// - On X11, the window must be visible.
+    pub fn fetch(&mut self) -> Result<Vec<u32>, SoftBufferError> {
+        self.surface_impl.fetch()
+    }
+
     /// Return a [`Buffer`] that the next frame should be rendered into. The size must
     /// be set with [`Surface::resize`] first. The initial contents of the buffer may be zeroed, or
     /// may contain a previous frame.
@@ -382,11 +391,6 @@ impl<'a> Buffer<'a> {
     /// Wayland compositor before calling this function.
     pub fn present(self) -> Result<(), SoftBufferError> {
         self.buffer_impl.present()
-    }
-
-    /// Copies the window contents into this buffer.
-    pub fn fetch(&mut self) -> Result<(), SoftBufferError> {
-        self.buffer_impl.fetch()
     }
 }
 
--- a/tests/present_and_fetch.rs
+++ b/tests/present_and_fetch.rs
@@ -1,0 +1,39 @@
+use softbuffer::{Context, Surface};
+use std::num::NonZeroU32;
+
+use std::panic::{catch_unwind, AssertUnwindSafe};
+
+use winit::event::Event;
+use winit::event_loop::{EventLoop, EventLoopWindowTarget};
+
+fn all_red(elwt: &EventLoopWindowTarget<()>) {
+    let window = winit::window::WindowBuilder::new()
+        .with_title("all_red")
+        .with_visible(false)
+        .build(elwt)
+        .unwrap();
+
+    let context = unsafe { Context::new(elwt) }.unwrap();
+    let mut surface = unsafe { Surface::new(&context, &window) }.unwrap();
+    let size = window.inner_size();
+
+    // Set the size of the surface to the size of the window.
+    surface.resize(
+        NonZeroU32::new(size.width).unwrap(),
+        NonZeroU32::new(size.height).unwrap(),
+    ).unwrap();
+
+    // Set all pixels to red.
+    let mut buffer = surface.buffer_mut().unwrap();
+    buffer.fill(0xFF0000FF);
+    buffer.present().unwrap();
+
+    // Check that all pixels are red.
+    let mut buffer = surface.buffer_mut().unwrap();
+    buffer.fetch().unwrap();
+    for pixel in buffer.iter() {
+        assert_eq!(*pixel, 0xFF0000FF);
+    }
+}
+
+winit_test::main!(all_red);
--- a/tests/present_and_fetch.rs
+++ b/tests/present_and_fetch.rs
@@ -1,38 +1,51 @@
 use softbuffer::{Context, Surface};
 use std::num::NonZeroU32;
-
-use std::panic::{catch_unwind, AssertUnwindSafe};
-
-use winit::event::Event;
-use winit::event_loop::{EventLoop, EventLoopWindowTarget};
+use winit::event_loop::EventLoopWindowTarget;
 
 fn all_red(elwt: &EventLoopWindowTarget<()>) {
     let window = winit::window::WindowBuilder::new()
         .with_title("all_red")
-        .with_visible(false)
         .build(elwt)
         .unwrap();
+
+    #[cfg(target_arch = "wasm32")]
+    {
+        use winit::platform::web::WindowExtWebSys;
+
+        web_sys::window()
+            .unwrap()
+            .document()
+            .unwrap()
+            .body()
+            .unwrap()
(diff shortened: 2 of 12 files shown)


ids1024:
On Wayland, if we want to read from the back buffer we can just do that. If we want to get window contents that this softbuffer `Surface` didn't present, there's no way to do so. For macOS and WASM presenting currently still copies, so this could just be a no-op. On Redox, presumably mapping the window and reading will work.

I'm not sure exactly how this API would be used though. It doesn't seem ideal to use every frame. If we have make a request to the display server and copy all the pixels to do this, would to be better for the application that wants to retain graphics to just have it's own persistent buffer and copy into the `softbuffer::Buffer` when it draws?

daxpedda:
Web implementation: https://github.com/rust-windowing/softbuffer/pull/105.

> If we want to get window contents that this softbuffer `Surface` didn't present, there's no way to do so.

I assumed that this is the purpose of this API. I implemented the Web implementation accordingly, otherwise it would be no-op like you suggested.

daxpedda:
Additionally, I'm unsure how other platforms then Web handle this, but what if the surface was resized, what does fetch do in this case, error? Or just get part of the contents that fit into the buffer? Does it zero out the rest if the fetched content is smaller then the current size?

ids1024:
> I assumed that this is the purpose of this API. 

If so, this isn't possible on Wayland. I'm guessing it's not possible on macOS given how the APIs work, but I would be less confident there.

It's not necessarily bad to expose features that only work on some backends, but I'd like to see more clearly what the actual application for this would be. And whether or not its the best way to achieve that.

notgull:
The goal of this API is to get the currently presented contents of the buffer and download it to the client. This is nice for testing, since it allows you to get the image created by the user. It would be nice for testing not only `softbuffer` but also drawing frameworks like [`theo`](https://github.com/notgull/theo).

ids1024:
Ah. For debugging uses it makes more sense. Then the performance concerns that would otherwise apply here aren't so much of a problem.

It should be clearly documented what this is and is not suitable for though.

I also wonder whether a method on `Buffer` like this is best. For debugging purposes, a `Surface` method returning `Result<Vec<u32>, _>` could be more suitable. If it's updating the contents of the `Buffer` that would clash with incremental rendering using buffer age when https://github.com/rust-windowing/softbuffer/pull/99 is merged.

notgull:
> I also wonder whether a method on `Buffer` like this is best. For debugging purposes, a `Surface` method returning `Result<Vec<u32>, _>` could be more suitable. If it's updating the contents of the `Buffer` that would clash with incremental rendering using buffer age when #99 is merged.

That's probably a good idea. I've changed this PR to use this design.

tests/present_and_fetch.rs
```diff
@@ -0,0 +1,79 @@
...
+    surface
+        .resize(
+            NonZeroU32::new(size.width).unwrap(),
+            NonZeroU32::new(size.height).unwrap(),
+        )
+        .unwrap();
+
+    // Set all pixels to red.
+    let mut buffer = surface.buffer_mut().unwrap();
+    buffer.fill(0xFF0000FF);
```
daxpedda, line 35: That looks like blue to me.
https://github.com/rust-windowing/softbuffer/blob/fcda747ddfb7062606787a369540db5badf03503/src/lib.rs#L340
```suggestion
    buffer.fill(0x00FF0000);
```

daxpedda:
> The goal of this API is to get the currently presented contents of the buffer and download it to the client.

So should this account for stuff presented not by `softbuffer`? Because if not, backends that don't have a buffer age can just return their buffer (if I understand this correctly).

notgull:
> > The goal of this API is to get the currently presented contents of the buffer and download it to the client.
> 
> So should this account for stuff not presented by `softbuffer`? Because if not, backends that don't have a buffer age can just return their buffer (if I understand this correctly).

Yes, this is what I wanted. Right now, my goal is to have a reproducible test harness; not only in `softbuffer`, but in other crates too. For `theo`, what I want to do is:

- Render an image to the window.
- Use `softbuffer` to fetch that image.
- Compare that image against an existing image to detect regressions.

tests/present_and_fetch.rs
ids1024: `0x00FF0000` won't be the same color on big-endian systems, right? If we care about those.

Something like `u32::from_ne_bytes` could be used for consistent behavior.

ids1024:
We'll need to consider how this interacts with https://github.com/rust-windowing/softbuffer/issues/98. Does the method `fetch` accept a format argument? Would the supported formats be the same as for presenting? Or does it return a value tagged with the format the surface contains? I'm not sure exactly how it works in the backends here.

Conversion is not necessarily lossless, if we support pixel formats that aren't just orderings RGB (though again, things like 10 bit color won't work with `u32` pixels).

https://github.com/rust-windowing/softbuffer/issues/98 is probably the most important feature missing in softbuffer currently, though I'm still not sure how best to design that, including conversion functionality.

tests/present_and_fetch.rs
daxpedda: I don't think so? I went down a rabbit hole for a bit there and confirmed two things:
1. Literals are always written in big-endian in Rust. I couldn't find any documentation on that, so I tried it out: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c1a0820eba008226bae7771c1f510fe8.
2. Bitshift operators always act in big-endian (except on some old PowerPC architectures I read somewhere?).

So I think what we are doing in Rust right now, seems to be pretty much cross-endianness compatible.
What I don't know is if the APIs expect us to turn things around on big-endian targets.
I also assume nobody here actually has a big-endian system lying around to try it out?

That said, on Web, my purview, this isn't a concern.

daxpedda:
> * Render an image to the window.
> * Use `softbuffer` to fetch that image.

On Web, the `Context` created has to be the same, if the original drawer used `"webgl2"`, `softbuffer` will fail, because it's using `"2d"`. Not sure if we care about this case, considering we are "**soft**buffer" and all.

Same if `transferControlToOffscreen()` was used, though this can be worked around with #103.

tests/present_and_fetch.rs
ids1024: I think the use of bitshifts in the examples would also be wrong on big endian. We represent color with `u32`, but it's really interpreted more like `[u8; 4]` but with 4 byte alignment. Which isn't dependent on endianess.

I don't have a big endian system to test on, though I fixed a similar issue in tiny-skia that someone reported: https://github.com/RazrFalcon/tiny-skia/pull/69. That one is easy to test with virtualization, but this is a bit more complicated since one may want a full OS with a display server. Maybe X11 backend over TCP could work...

daxpedda: > Maybe X11 backend over TCP could work...

Sounds like fun, please share the result when you get them :rofl:.

daxpedda: > 2\. Bitshift operators always act in big-endian

> I think the use of bitshifts in the examples would also be wrong on big endian.

I think these two contradict each other, if bitshift operators are always acting in big-endian, then it shouldn't matter what platform you are on. Unless, looking at your PR, you were using PowerPC to test: 
> (except on some old PowerPC architectures I read somewhere?)

Or maybe my assumptions are just wrong, I can't really test them.

ids1024: I don't think "bitshift operators are always acting in big-endian" really makes sense? That would mean `1 << 3` for instance isn't `8`, because when compared in native endianess it is interpreted as something other than `8`?

It looks like you're right that there's something weird about how bitshifts and endianness work on powerpc, at the assembly level, but at the Rust or C level they should behave the same way across platforms in how they operate on integer values, while the actual order of the bytes in memory may differ.

daxpedda: > I don't think "bitshift operators are always acting in big-endian" really makes sense? That would mean `1 << 3` for instance isn't `8`, because when compared in native endianess it is interpreted as something other than `8`?

No, what I'm saying is that `1 << 3` is always `8`, no matter the endianness of the platform. That is why I think any code we are using here should be right.

... unless of course the API isn't really seeing a `u32`, but a `[u8; 4]`, in which case I would think that the same number would have a different meanings depending on the platform endianness.

> We represent color with `u32`, but it's really interpreted more like `[u8; 4]` but with 4 byte alignment. Which isn't dependent on endianess.

... contrary to what you said earlier.

(lots of emphases on "I think", again: can't test any of my assumptions here)

ids1024: I mean that the pixel format is basically a `[u8; 4]`, or "big-endian" BGRX, (B byte, G byte, R byte, then X byte) regardless of of platform endianess. So the *byte order* of the pixels doesn't depend on endianness, but consequently the integer value this corresponds to will depend on endianess.

I guess to test this I'll need to cross compile at least libxcb for big endinan powerPC or mips. I haven't had luck getting either to boot in a VM, though if I can cross compile a binary qemu user emulation should work...

But this isn't that important if it's hard to even find something big endian to emulate. It would just be nice to do the correct thing here.

daxpedda: > I mean that the pixel format is basically a `[u8; 4]`, or "big-endian" BGRX, (B byte, G byte, R byte, then X byte) regardless of of platform endianess. So the _byte order_ of the pixels doesn't depend on endianness, but consequently the integer value this corresponds to will depend on endianess.

This sounds very correct to me.

So yeah, agreed, this would need correction.

---

On that note, why did we decide to use `u32` instead of `[u8; 4]`?

ids1024: It can be faster to work with 32-bit values than individual bytes. And things like tiny-skia work on `&[u32]`. And you can safely cast `&[u32]` to `&[u8]` but not the other way around due to alignment.

daxpedda: I'm not sure what to think of this, there isn't actually a **safe** method to cast `&[u32]` to `&[u8]` I know of. I would argue that `u32` is definitely a worse user interface and the improved speed is speculation (I actually agree with that speculation).

Just an idea: what do you think of `&[[u8; 4]]`?

ids1024: `[u32]` has alignment `4` (on typical architectures), while `[u8; 4]` has alignment `1`. So there's not currently a standard library method to cast `&[u32]` to `&[u8]`, but it's always sound while the reverse is not true. Likewise it is potentially unsound to cast `&[[u8; 4]]` to `&[u32]`.

`bytemuck` is popular for this sort of thing.

daxpedda: > while `[u8; 4]` has alignment `1`.

Ah, apologies, was thinking that this does have an alignment of 4 actually, for some reason (I think I played around with too much shaders lately ...).

---

Alright, w/e, I think I'm still in favor of `&[u8]` until we can show a reasonable performance degradation, but this is an issue for another time.

In the meantime, your original problem still needs to be addressed. I think we should also document this somewhere.
https://github.com/rust-windowing/softbuffer/blob/fcda747ddfb7062606787a369540db5badf03503/src/lib.rs#L340
This could be a bit confusing, because it's showing big-endian order, which most platforms aren't, but it's also representative of how literals are written in Rust.

daxpedda: I say we handle it afterwards: https://github.com/rust-windowing/softbuffer/issues/109.

notgull:
I think this is ready now. macOS, Wayland and Redox are unimplemented for now; I'm hoping that the platform maintainers will be able to take care of that.

notgull:
I'm unsure why WASM is failing by returning an empty buffer, nor am I sure why Windows seems to spurious fail in the same way.

daxpedda:
It's because https://github.com/rust-windowing/softbuffer/pull/104#discussion_r1193109522 wasn't applied. Currently fails on X11 because of that too.

src/lib.rs
```diff
@@ -306,6 +315,16 @@ impl Surface {
         self.surface_impl.resize(width, height)
     }
 
+    /// Copies the window contents into a buffer.
+    ///
+    /// ## Platform Dependent Behavior
```
ids1024, line 320: It was mentioned in https://github.com/rust-windowing/softbuffer/pull/104#issuecomment-1546966584 that this won't work for "window" contents set from webgl. Sounds like a caveat that should be mentioned here? If the intention is to capture window contents not presented with softbuffer.

Not sure how that works on Windows or X11.

daxpedda: It's kinda covered in the documentation here, saying an object "valid to draw a 2D buffer to":
https://github.com/rust-windowing/softbuffer/blob/eb227f10f06f8e5cd4d34ab0c3694c4e60606ff7/src/lib.rs#L226-L227
And throws the appropriate error here:
https://github.com/rust-windowing/softbuffer/blob/eb227f10f06f8e5cd4d34ab0c3694c4e60606ff7/src/web.rs#L68-L70

Though I'm a big fan of https://rust-lang.github.io/rust-clippy/master/index.html#missing_errors_doc and would prefer us to document every possible error on every function.

I made https://github.com/rust-windowing/softbuffer/pull/113 and added some documentation where I felt it was appropriate.

notgull:
Looks like adding `GdiFlush` fixed the Win32 backend. Hopefully this means that all of the outstanding issues are resolved.

ids1024 (approved):

PR #105, opened 2023-05-13
Add `fetch()` Web Implementation

daxpedda:
Web implementation for https://github.com/rust-windowing/softbuffer/pull/104.

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -44,6 +44,9 @@ pub struct WebImpl {
 
     /// The current width of the canvas.
     width: u32,
+
+    /// The current height of the canvas.
+    height: u32,
 }
 
 impl WebImpl {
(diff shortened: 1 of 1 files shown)


daxpedda:
The test introduced in #104 isn't compatible with Wasm btw, because of `inventory`. I'm unsure of the motivations here, but it seems to me that this could be simplified.

Alternatively we can just disable the test for Wasm and I write a separate one for Wasm only.

notgull (approved):
Thank you!

> The test introduced in #104 isn't compatible with Wasm btw, because of `inventory`. I'm unsure of the motivations here, but it seems to me that this could be simplified.

Drat. Maybe we could hammer `linkme` into a usable state instead?

PR #107, opened 2023-05-14
Adjust `fetch()` test for Web

daxpedda:
Had to butcher the test a bit, apologies.


notgull (approved):
Thank you!

It might be useful to extract some of this logic into a `winit-test` crate that uses a centralized `EventLoop` to run tests.

PR #111, opened 2023-05-27
Bump `x11rb` to v0.12

daxpedda:
Should have no breaking changes for us: https://github.com/psychon/x11rb/blob/v0.12.0/doc/changelog.md?plain=1#L1C1-L31.


notgull (approved):
Thanks!

PR #116, opened 2023-06-02
Remove thiserror as a dependency

notgull:
We can pretty easily just implement its tricks for ourselves without needing to import `syn`.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,50 +2,139 @@ use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 use std::error::Error;
 use std::fmt;
 use std::num::NonZeroU32;
-use thiserror::Error;
 
-#[derive(Error, Debug)]
-#[allow(missing_docs)] // TODO
+#[derive(Debug)]
 #[non_exhaustive]
+/// A sum type of all of the errors that can occur during the operation of this crate.
 pub enum SoftBufferError {
-    #[error(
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -2,44 +2,128 @@ use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
...
+            Self::IncompleteWindowHandle => write!(f, "The provided window handle is null."),
+            Self::IncompleteDisplayHandle => write!(f, "The provided display handle is null."),
+            Self::SizeOutOfRange { width, height } => write!(
+                f,
+                "Surface size {}x{} out of range for backend.",
+                width, height
+            ),
+            Self::PlatformError(msg, None) => write!(f, "Platform error: {:?}", msg),
+            Self::PlatformError(msg, Some(err)) => write!(f, "Platform error: {:?}: {}", msg, err),
+        }
```
daxpedda, line 123: We can already use implicit named arguments (`format!("{msg}")`). Was stabilized in 1.58, our MSRV is 1.64.

daxpedda (changes requested):
This drops `syn` for all targets except Android and Wasm, amazing!

src/error.rs
notgull: I used implicit named arguments for the shorter names. I think that the longer names shouldn't be embedded into the format string.

daxpedda (approved):

PR #118, opened 2023-06-04
v0.3.0

notgull:
* On MacOS, the contents scale is updated when set_buffer() is called, to adapt when the window is on a new screen (#68).
* **Breaking:** Split the `GraphicsContext` type into `Context` and `Surface` (#64).
* On Web, cache the document in the `Context` type (#66).
* **Breaking:** Introduce a new "owned buffer" for no-copy presentation (#65).
* Enable support for multi-threaded WASM (#77).
* Fix buffer resizing on X11 (#69).
* Add a set of functions for handling buffer damage (#99).
* Add a `fetch()` function for getting the window contents (#104).
* Bump MSRV to 1.64 (#81).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
-# UNRELEASED
+# 0.3.0
 
-* On MacOS, the contents scale is updated when set_buffer() is called, to adapt when the window is on a new screen.
+* On MacOS, the contents scale is updated when set_buffer() is called, to adapt when the window is on a new screen (#68).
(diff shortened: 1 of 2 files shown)


notgull:
I don't think that I've missed anything; I also don't think any of the pending issues are so bad that they should stop a new release.

cc [this](https://github.com/rust-windowing/winit/pull/2812#issuecomment-1575710393)

ids1024 (approved):

PR #125, opened 2023-06-14
Make documentation clearer

StratusFearMe21:
From your documentation, I was under the impression that the pixel format of `softbuffer` was `XRGB`, however, as a maintainer of iced-rs/iced told me in [this](https://github.com/iced-rs/iced/discussions/1915#discussioncomment-6175061) discussion, the buffer format is actually `BGRX`. I think this documentation change would make that clearer.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -378,7 +378,7 @@ impl Surface {
 ///
 /// # Data representation
 ///
-/// The format of the buffer is `BGRX8888`, which is as follows. There is one `u32` in the buffer for each pixel in
+/// The format of the buffer is `BGR0`, which is as follows. There is one `u32` in the buffer for each pixel in
 /// the area to draw. The first entry is the upper-left most pixel. The second is one to the right
 /// etc. (Row-major top to bottom left to right one `u32` per pixel). Within each `u32` the highest
 /// order 8 bits are to be set to 0. The next highest order 8 bits are the red channel, then the
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -378,7 +378,7 @@ impl Surface {
 ///
 /// # Data representation
 ///
-/// The format of the buffer is as follows. There is one `u32` in the buffer for each pixel in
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -378,7 +378,7 @@ impl Surface {
 ///
 /// # Data representation
 ///
-/// The format of the buffer is as follows. There is one `u32` in the buffer for each pixel in
+/// The format of the buffer is `BGRX8888`, which is as follows. There is one `u32` in the buffer for each pixel in
```
daxpedda, line 381:
```suggestion
/// The format of the buffer is `BGR0`, which is as follows. There is one `u32` in the buffer for each pixel in
```
I'm not versed in these kind of terms, is there some precedence for naming it like you did?

StratusFearMe21: It's based on how Wayland formats are described, but I like your description better

kpreid:
As written, this PR makes the documentation self-contradictory. Right now, the documentation says that

> Pixel format (u32):
> 
> 00000000RRRRRRRRGGGGGGGGBBBBBBBB

By saying this is a u32 and displaying bits, this to me means that the least significant byte is the blue component and the most significant byte is zero. That means that the _byte_ ordering of the pixel data is [B,G,R,0] on little-endian machines and [0,R,G,B] on big-endian machines. Claiming that it is BGR0 always would contradict that.

Not having tested softbuffer on a big-endian machine, I don't know which is correct, but the documentation should not contradict itself. (And if the byte ordering is BGR0 regardless of endianness, then the API should represent that as `[u8; 4]`, not `u32`!)

StratusFearMe21:
Oooohhhh. I get it now. I didn't realize that's how that worked

daxpedda:
> And if the byte ordering is BGR0 regardless of endianness, then the API should represent that as `[u8; 4]`, not `u32`!)

We suspect that this is actually the case and I'm in favor of changing the API to be `[u8; 4]`.

PR #126, opened 2023-06-15
Use tiny-xlib instead of x11-dl

notgull:
As of now, `x11-dl` takes up a significant part of our executable.

image

In order to avoid this bloat, especially since we only really use one function from Xlib, I've created the [`tiny-xlib`](https://crates.io/crates/tiny-xlib) crate to act as a small interface over Xlib and Xlib-Xcb. This PR replaces the `x11-dl` crate with `tiny-xlib`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
           - { target: i686-pc-windows-gnu,      os: windows-latest, host: -i686-pc-windows-gnu }
           - { target: i686-unknown-linux-gnu,   os: ubuntu-latest,   }
           - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest,   }
-          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest, options: --no-default-features, features: x11 }
+          - { target: x86_64-unknown-linux-gnu, os: ubuntu-latest, options: --no-default-features, features: "x11,x11-dlopen" }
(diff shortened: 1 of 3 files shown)


camtauxe:
This patch breaks any executables using this crate on NixOS, since the tiny-xlib library is unable to find libX11 library, even if it's installed by including it in `buildInputs`.

In my `nix-shell` environment, I was able to get around this by setting the `LD_LIBRARY_PATH` variable like so:
```nix
with pkgs; LD_LIBRARY_PATH = lib.makeLibraryPath [ xorg.libX11 ]
```

This allows me to resume development on my own crate, but I don't know how I will be able to distribute it in any usable form now.

notgull:
Thanks for finding this out! Please file an issue on the `tiny-xlib` repo

Ralith:
@camtauxe NixOS binary packages are routinely distributed with wrapper scripts and/or `patchelf` fixups to support this type of `dlopen` use, but you can probably also just disable the `x11-dlopen` feature and put `xorg.libX11` in `buildInputs` like usual. Not sure why this patch changed anything for you, since it looks like it was using dlopen beforehand anyway?

camtauxe:
Apologies for the sour tone in my last comment, I appreciate the advice! Disabling the `x11-dlopen` feature did the trick. And "`dlopen`" was the term I was missing when trying to describe the problem to Google, so I was able to find some help there. [This thread](https://discourse.nixos.org/t/how-to-set-runtimedependencies-for-shared-libraries/14857) on the NixOS discourse in particular has a variety of suggestions.

PR #127, opened 2023-06-16
examples/winit: Implement proper `resumed()` semantics

MarijnS95:
On Android the backing buffer (`NativeWindow`) disappears when the application is not focussed and/or the screen is locked.  Winit handles this by requiring apps to create their `raw_window_handle()` consumers _after_ `Event::Resumed` and to clean it up _before_ returning from `Event::Suspended`.  For consistency Winit also sends `Resumed` on all other platforms during init.

---

Draft until this is transplanted to all the other examples, even though that doesn't magically make them run on Android yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,21 +79,27 @@ mod winit_app;
 fn main() {
     let event_loop = EventLoop::new().unwrap();
 
-    let mut app = winit_app::WinitAppBuilder::with_init(|elwt| {
-        let window = {
-            let window = elwt.create_window(Window::default_attributes());
-            Rc::new(window.unwrap())
-        };
-        let context = softbuffer::Context::new(window.clone()).unwrap();
-        let surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+    let mut app = winit_app::WinitAppBuilder::with_init(
+        |elwt| {
+            let window = {
+                let window = elwt.create_window(Window::default_attributes());
+                Rc::new(window.unwrap())
+            };
+            let context = softbuffer::Context::new(window.clone()).unwrap();
 
-        (window, surface)
-    }).with_event_handler(|state, event, elwt| {
-        let (window, surface) = state;
+            (window, context)
+        },
+        |_elwt, (window, context)| softbuffer::Surface::new(context, window.clone()).unwrap(),
+    )
+    .with_event_handler(|(window, _context), surface, event, elwt| {
         elwt.set_control_flow(ControlFlow::Wait);
 
         match event {
             Event::WindowEvent { window_id, event: WindowEvent::RedrawRequested } if window_id == window.id() => {
+                let Some(surface) = surface else {
+                    eprintln!("RedrawRequested fired before Resumed or after Suspended");
+                    return;
+                };
                 let (width, height) = {
                     let size = window.inner_size();
                     (size.width, size.height)
--- a/examples/utils/winit_app.rs
+++ b/examples/utils/winit_app.rs
@@ -31,76 +31,94 @@ pub(crate) fn make_window(
 }
 
 /// Easily constructable winit application.
-pub(crate) struct WinitApp<T, Init, Handler> {
-    /// Closure to initialize state.
+pub(crate) struct WinitApp<T, S, Init, InitSurface, Handler> {
+    /// Closure to initialize `state`.
     init: Init,
+
+    /// Closure to initialize `surface_state`.
+    init_surface: InitSurface,
 
     /// Closure to run on window events.
     event: Handler,
 
     /// Contained state.
     state: Option<T>,
+
+    /// Contained surface state.
+    surface_state: Option<S>,
 }
 
 /// Builder that makes it so we don't have to name `T`.
-pub(crate) struct WinitAppBuilder<T, Init> {
-    /// Closure to initialize state.
+pub(crate) struct WinitAppBuilder<T, S, Init, InitSurface> {
+    /// Closure to initialize `state`.
     init: Init,
 
+    /// Closure to initialize `surface_state`.
+    init_surface: InitSurface,
+
     /// Eat the type parameter.
-    _marker: PhantomData<Option<T>>,
+    _marker: PhantomData<(Option<T>, Option<S>)>,
 }
 
-impl<T, Init> WinitAppBuilder<T, Init>
+impl<T, S, Init, InitSurface> WinitAppBuilder<T, S, Init, InitSurface>
 where
     Init: FnMut(&ActiveEventLoop) -> T,
+    InitSurface: FnMut(&ActiveEventLoop, &mut T) -> S,
 {
     /// Create with an "init" closure.
-    pub(crate) fn with_init(init: Init) -> Self {
+    pub(crate) fn with_init(init: Init, init_surface: InitSurface) -> Self {
         Self {
             init,
+            init_surface,
             _marker: PhantomData,
         }
     }
 
     /// Build a new application.
-    pub(crate) fn with_event_handler<F>(self, handler: F) -> WinitApp<T, Init, F>
+    pub(crate) fn with_event_handler<F>(self, handler: F) -> WinitApp<T, S, Init, InitSurface, F>
     where
-        F: FnMut(&mut T, Event<()>, &ActiveEventLoop),
+        F: FnMut(&mut T, Option<&mut S>, Event<()>, &ActiveEventLoop),
     {
-        WinitApp::new(self.init, handler)
+        WinitApp::new(self.init, self.init_surface, handler)
     }
 }
 
-impl<T, Init, Handler> WinitApp<T, Init, Handler>
+impl<T, S, Init, InitSurface, Handler> WinitApp<T, S, Init, InitSurface, Handler>
 where
     Init: FnMut(&ActiveEventLoop) -> T,
-    Handler: FnMut(&mut T, Event<()>, &ActiveEventLoop),
+    InitSurface: FnMut(&ActiveEventLoop, &mut T) -> S,
+    Handler: FnMut(&mut T, Option<&mut S>, Event<()>, &ActiveEventLoop),
 {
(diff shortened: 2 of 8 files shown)


notgull:
This was done as a part of #314, so I'll close this PR

MarijnS95:
@notgull I've started reworking this patch into https://github.com/MarijnS95/softbuffer/compare/examples-resumed-semantics, let me know what you think before I start reworking the other examples and PR.

Note that there is some confusion currently in the `winit` APIs that we're cleaning up upstream in https://github.com/rust-windowing/winit/pull/3765 and https://github.com/rust-windowing/winit/issues/3779.

MarijnS95:
Oh @notgull if you can reopen I can just (force)push the rebase/rework here and we can keep all this discussion context :)

examples/winit_multithread.rs
```diff
@@ -59,41 +63,55 @@ mod ex {
...
+
+                // Spawn a thread to handle rendering.
+                let (start_render, do_render) = mpsc::channel();
+                let (render_done, finish_render) = mpsc::channel();
+                println!("starting thread...");
+                std::thread::spawn({
+                    let window = window.clone();
+                    let surface = surface.clone();
+                    move || render_thread(window, surface, do_render, render_done)
+                });
```
MarijnS95, line 97: Thoughts?

MarijnS95:
Note that, to actually implement the Android backend, there are ideas to (temporarily?) provide the user a fake `Vec` without stride and the current fixed pixel format. This buffer will be "transformed/converted" into the Android buffer on "submit".

If we do that, we could also pretend to the user that their `Surface` "always" exist (and give them a buffer), and handle the lifetime internally.

examples/utils/winit_app.rs
```diff
@@ -1,122 +1,151 @@
...
-            event: WindowEvent,
-        ) {
-            let state = self.state.as_mut().unwrap();
-            (self.event)(state, Event::WindowEvent {
-                window_id,
-                event
-            }, event_loop);
-        }
+    fn suspended(&mut self, _event_loop: &ActiveEventLoop) {
+        // TODO: Should we run a destruction function for the user?
```
daxpedda, line 119: We are already destroying the surface, is there more to do here or what do you mean with "for the user"?

examples/winit_multithread.rs
daxpedda: I actually think its fine like this.
The thread is destroyed on `suspended()` anyway, maybe we should leave a comment stating as much.

Probably update this line as well:
https://github.com/rust-windowing/softbuffer/blob/592c49e4876979bc09b8f5002edcb9b345f6c9c5/examples/winit_multithread.rs#L29

README.md
```diff
@@ -69,7 +69,8 @@ use winit::event::{Event, WindowEvent};
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::Window;
 
-include!("../examples/utils/winit_app.rs");
+#[path = "../examples/utils/winit_app.rs"]
```
daxpedda, line 72: I would have gone further and not used `path` at all.
This can be done by moving the file directly to `examples/winit_app.rs` and excluding it as an auto discoverable example with `[[example]]` in `Cargo.toml`.

Just a suggestion, can be done in a follow-up as well or not at all if ppl don't like it.

daxpedda (commented):
LGTM!
CI is broken because of the changes in `WinitApp` right?

README.md
MarijnS95: @daxpedda this was handled in https://github.com/rust-windowing/softbuffer/pull/227, I don't remember if I asked whether to add a `mod.rs` instead.

examples/utils/winit_app.rs
MarijnS95: We implicitly run `Drop` on it (and the `softbuffer::Surface` that we hold implements proper semantics), I don't think there's anything for the user to handle in here.

It's example code anyway, if users want to build off of this and need that functionality they can always add it themselves.

examples/winit_multithread.rs
MarijnS95: Okay, I've updated this a bit but would like to test it, including the idea to keep one thread alive for the context and sending it the `Surface` as part of the app (not surface) state.

daxpedda (approved):
LGTM!
Would like to replace `(e)print(ln)` with `tracing`, but that's for a follow-up.

Documentation tests have to be fixed.

examples/winit_multithread.rs
MarijnS95: Forgot to mention that I still wanted to test this on an actual Android phone, both making sure that the current logic works and the idea I had about persistent threads.

daxpedda:
@MarijnS95 rebasing on #228 should fix the remaining CI.

examples/winit_multithread.rs
MarijnS95:
```
08-03 11:15:00.827 15549 15549 D android_activity::act..: NativeWindowCreated: 0xb400007229e9ff70 -- 0xb400007369e7ef10
08-03 11:15:00.827 15549 15549 D android_activity::act..: NativeWindowResized: 0xb400007229e9ff70 -- 0xb400007369e7ef10
08-03 11:15:00.828 15549 15598 I RustStdoutStderr: making surface...
08-03 11:15:00.828 15549 15598 I RustStdoutStderr: starting thread...
08-03 11:15:00.829 15549 15598 I RustStdoutStderr: waiting for render...
08-03 11:15:00.829 15549 15549 D android_activity::act..: ContentRectChanged: 0xb400007229e9ff70 -- 0x7fded9b708
08-03 11:15:00.831 15549 15549 D android_activity::act..: NativeWindowRedrawNeeded: 0xb400007229e9ff70 -- 0xb400007369e7ef10
08-03 11:15:00.832 15549 15598 I RustStdoutStderr: got size: PhysicalSize { width: 1096, height: 2560 }
08-03 11:15:00.832 15549 15598 I RustStdoutStderr: resizing...
08-03 11:15:00.838 15549 15603 I AdrenoUtils: <ReadGpuID:408>: Reading chip ID through GSL
08-03 11:15:00.851 15549 15549 D android_activity::act..: WindowFocusChanged: 0xb400007229e9ff70 -- 1
08-03 11:15:01.087 15549 15598 I RustStdoutStderr: presenting...
08-03 11:15:01.087 15549 15598 I RustStdoutStderr: waiting for render...
08-03 11:15:01.106 15549 15575 I ithread_android: Compiler allocated 4265KB to compile void android.widget.TextView.<init>(android.content.Context, android.util.AttributeSet, int, int)
08-03 11:15:01.165 15549 15575 I ithread_android: Compiler allocated 4683KB to compile void android.view.ViewRootImpl.performTraversals()
08-03 11:15:03.577 15549 15549 D android_activity::act..: Pause: 0xb400007229e9ff70
08-03 11:15:03.611 15549 15549 D android_activity::act..: NativeWindowDestroyed: 0xb400007229e9ff70 -- 0xb400007369e7ef10
08-03 11:15:03.612 15549 15598 I RustStdoutStderr: Main thread destroyed
```
Works as it should on Android!

(note that I added `Main thread destroyed` in `do_render.recv().is_err()` even though that is not really what it does currently, but it shows that destruction of the surface state and the relevant `mpsc` is happening correctly.)

daxpedda: > (note that I added `Main thread destroyed` in `do_render.recv().is_err()` even though that is not really what it does currently, but it shows that destruction of the surface state and the relevant `mpsc` is happening correctly.)

You corrected it to:
https://github.com/rust-windowing/softbuffer/blob/eb6be27e0284a93ec613cdc95dd6bc4aee724fdb/examples/winit_multithread.rs#L29

So looks good, right?

MarijnS95: I meant in the example log above.

I've ultimately rectified it in both commits in the history, as I ended up pushing this extra log line for testing.

examples/winit_multithread.rs
```diff
@@ -19,16 +19,16 @@ mod ex {
...
-        do_render: mpsc::Receiver<()>,
+        do_render: mpsc::Receiver<Arc<Mutex<Surface>>>,
         done: mpsc::Sender<()>,
     ) {
         loop {
             println!("waiting for render...");
-            if do_render.recv().is_err() {
+            let Ok(surface) = do_render.recv() else {
+                println!("main thread destroyed");
                 // Main thread is dead.
```
daxpedda, line 29: I don't think the comment is needed if we print a line saying the same thing as well that is very human readable.

daxpedda (approved):

MarijnS95:
Only nightly seems to fail on `#[path = ...] mod winit_app;`, perhaps irrelevant to this PR?

MarijnS95:
Yikes, this relative path error, that we first [only saw on `nightly`](https://github.com/rust-windowing/softbuffer/actions/runs/11235857449), is now [only showing on `stable`](https://github.com/rust-windowing/softbuffer/actions/runs/12090860880):

```
---- src/../README.md - (line 69) stdout ----
error: couldn't read src/../../examples/utils/winit_app.rs: No such file or directory (os error 2)
Error:   --> src/../README.md:78:1
   |
11 | mod winit_app;
   | ^^^^^^^^^^^^^^

error: aborting due to 1 previous error

Couldn't compile the test.

failures:
    src/../README.md - (line 69)

test result: FAILED. 2 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
```

Did the fix (assuming it is a fix to how relative paths from content in `include_str!()` are resolved) get reverted, removed, or regressed on `nightly` again?

After all, I'd argue the path should be relative to `README.md` no matter what, not to wherever this file in `#[doc = include_str!()]`'d which could be at any arbitrary level in our directory tree...

Pushed a "fix", let's see if nightly would then start to break again...

MarijnS95:
It breaks both stable and our MSRV... How do you want to fix this, @ids1024 @notgull?

PR #130, opened 2023-06-26
Android backend

### LINKED ISSUE #44: Android platform

MarijnS95:
Depends on #127
Fixes #44

Contains my hacks to further discussing Android support.  There are two issues:

- Format is RGBA or RGBX;
- Buffers almost always have a `stride() != width()`, requiring us to deal with that (and `IOSurface` on Mac seems to have the same?).

These could either be solved by exposing this information to the user (more work and documentation for them, but also more performant) or kept hidden by maintaining the current API, but requiring an extra framebuffer allocation that we convert when they submit the frame.

### DIFF
--- a/examples/winit.rs
+++ b/examples/winit.rs
@@ -6,9 +6,12 @@ use winit::keyboard::{Key, NamedKey};
 #[path = "utils/winit_app.rs"]
 mod winit_app;
 
+#[cfg(not(target_os = "android"))]
 fn main() {
-    let event_loop = EventLoop::new().unwrap();
+    entry(EventLoop::new().unwrap())
+}
 
+pub(crate) fn entry(event_loop: EventLoop<()>) {
     let app = winit_app::WinitAppBuilder::with_init(
         |elwt| {
             let window = winit_app::make_window(elwt, |w| w);
--- a/src/backends/android.rs
+++ b/src/backends/android.rs
@@ -1,0 +1,154 @@
+//! Implementation of software buffering for Android.
+
+use std::marker::PhantomData;
+use std::num::{NonZeroI32, NonZeroU32};
+
+use ndk::{
+    hardware_buffer_format::HardwareBufferFormat,
+    native_window::{NativeWindow, NativeWindowBufferLockGuard},
+};
+use raw_window_handle::{HasDisplayHandle, HasWindowHandle, RawWindowHandle};
+
+use crate::error::InitError;
+use crate::{Rect, SoftBufferError};
+
+/// The handle to a window for software buffering.
+pub struct AndroidImpl<D: ?Sized, W: ?Sized> {
+    native_window: NativeWindow,
+
+    _display: PhantomData<D>,
+
+    /// The pointer to the window object.
+    ///
+    /// This is pretty useless because it gives us a pointer to [`NativeWindow`] that we have to increase the refcount on.
+    /// Alternatively we can use [`NativeWindow::from_ptr()`] wrapped in [`std::mem::ManuallyDrop`]
+    window: W,
+}
+
+// TODO: Current system doesn't require a trait to be implemented here, even though it exists.
+impl<D: HasDisplayHandle, W: HasWindowHandle> AndroidImpl<D, W> {
+    /// Create a new [`AndroidImpl`] from an [`AndroidNdkWindowHandle`].
+    ///
+    /// # Safety
+    ///
+    /// The [`AndroidNdkWindowHandle`] must be a valid window handle.
+    // TODO: That's lame, why can't we get an AndroidNdkWindowHandle directly here
+    pub(crate) fn new(window: W, _display: &D) -> Result<Self, InitError<W>> {
+        // Get the raw Android window (surface).
+        let raw = window.window_handle()?.as_raw();
+        let RawWindowHandle::AndroidNdk(a) = raw else {
+            return Err(InitError::Unsupported(window));
+        };
+
+        // Acquire a new owned reference to the window, that will be freed on drop.
+        let native_window = unsafe { NativeWindow::clone_from_ptr(a.a_native_window.cast()) };
+
+        Ok(Self {
+            native_window,
+            // _display: DisplayHandle::borrow_raw(raw_window_handle::RawDisplayHandle::Android(
+            //     AndroidDisplayHandle,
+            // )),
+            _display: PhantomData,
+            window,
+        })
+    }
+
+    #[inline]
+    pub fn window(&self) -> &W {
+        &self.window
+    }
+
+    /// Also changes the pixel format to [`HardwareBufferFormat::R8G8B8A8_UNORM`].
+    pub fn resize(&mut self, width: NonZeroU32, height: NonZeroU32) -> Result<(), SoftBufferError> {
+        let (width, height) = (|| {
+            let width = NonZeroI32::try_from(width).ok()?;
+            let height = NonZeroI32::try_from(height).ok()?;
+            Some((width, height))
+        })()
+        .ok_or(SoftBufferError::SizeOutOfRange { width, height })?;
+
+        // Do not change the format.
+        self.native_window
+            .set_buffers_geometry(
+                width.into(),
+                height.into(),
+                // Default is typically R5G6B5 16bpp, switch to 32bpp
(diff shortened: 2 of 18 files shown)


ids1024:
> These could either be solved by exposing this information to the user (more work and documentation for them, but also more performant) or kept hidden by maintaining the current API, but requiring an extra framebuffer allocation that we convert when they submit the frame.

Currently the web backend converts to RGBA on present. We can do the same here, if we want to merge this before we figure out https://github.com/rust-windowing/softbuffer/issues/98.

Ultimately we'll want that to provide an API that lets a client enumerate supported formats (some backends may support only BGRA, only RGBA, or both, etc.). And either use a native format or provide an easy way to convert when needed.

ids1024:
Adding a `stride()` is unnecessary for now, if it converts on present anyway. But we'll need that for this for `IOSurface`, or anything else where we write directly into memory in a format the GPU may handle with a stride requirement.

So that can be added to the API when https://github.com/rust-windowing/softbuffer/issues/98 is implemented.

examples/winit.rs
```diff
@@ -1,10 +1,30 @@
 use std::num::NonZeroU32;
+
 use winit::event::{Event, WindowEvent};
-use winit::event_loop::{ControlFlow, EventLoop};
+use winit::event_loop::{ControlFlow, EventLoop, EventLoopBuilder};
+#[cfg(target_os = "android")]
+pub use winit::platform::android::{activity::AndroidApp, EventLoopBuilderExtAndroid};
 use winit::window::WindowBuilder;
 
+/// Run with `cargo apk r --example winit_android`
```
ids1024, line 9: We have a mention of how to run the example on webassembly in the readme: https://github.com/rust-windowing/softbuffer/#webassembly.

So a quick mention of Android there would be good.

MarijnS95:
> > These could either be solved by exposing this information to the user (more work and documentation for them, but also more performant) or kept hidden by maintaining the current API, but requiring an extra framebuffer allocation that we convert when they submit the frame.
> 
> Currently the web backend converts to RGBA on present. We can do the same here, if we want to merge this before we figure out #98.

No fuss/hurry to get this merged before fleshing out the details.

> Ultimately we'll want that to provide an API that lets a client enumerate supported formats (some backends may support only BGRA, only RGBA, or both, etc.). And either use a native format or provide an easy way to convert when needed.

Sounds like a plan.

> Adding a `stride()` is unnecessary for now, if it converts on present anyway. But we'll need that for this for `IOSurface`, or anything else where we write directly into memory in a format the GPU may handle with a stride requirement.

It's necessary unless we code in the conversion on present. And Android might again do a conversion after that.

examples/utils/winit_app.rs
MarijnS95, line 1: Any clue why this file used an `include!()` hack? Making it a module makes sure `cargo fmt` tackles it, and probably makes it easier to have jump to symbol and friends working properly.

ids1024: Not sure when that's  used, seems a little odd.

Looks like that was added by @notgull in 0e6706c7afba582deaa8c47ff049d03cb0f8e363.

notgull: You can't use modules in examples, I think

daxpedda: I think you can if you put it directly in the examples folder and [exclude it as an example via cargo targets](https://doc.rust-lang.org/1.79.0/cargo/reference/cargo-targets.html#target-auto-discovery).

(would have tried it out, but am on my phone right now)

MarijnS95: You can definitely use `mod`ules in examples. The only thing you have to do, which is already done here and as linked "somewhat" by @daxpedda, is to make sure that the utilities library doesn't get picked up by `autoexamples = true`. This is already done, because the only files that get picked up are `examples/<example name>.rs` and `examples/<example name>/main.rs`. As there's no `main.rs` inside the `utils/` folder, there's no `utils` example target.

src/backends/android.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /// The number of _pixels_ that a line in the buffer takes in memory.
+    #[inline]
+    pub fn stride(&self) -> u32 {
+        self.0.stride() as u32
+    }
+
+    pub fn age(&self) -> u8 {
+        todo!()
```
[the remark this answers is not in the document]
MarijnS95: Android works with buffer pools and buffer queues, to implement multi-buffering. The [underlying function](https://developer.android.com/ndk/reference/group/a-native-window#group___a_native_window_1ga0b0e3b7d442dee83e1a1b42e5b0caee6) is defined as:

> Lock the window's next drawing surface for writing.

Previous data is probably undefined, if a different handle is returned most of the time. However, I've implemented a `Vec` with a copy of the data, which we might use for this (and to implement the new `fn fetch()`?).

---

In a quick test I just kept track of the raw pointers returned from this function. After 3 calls the pointer repeats, and always in the same order. If I don't write to this pointer the old data is preserved.

I have no clue if this is something we can rely on however, since Android doesn't specify the lifetime of this CPU handle. And depending on the phone hardware, display presentation unit and UMA, may or may not have to copy this CPU visible memory (and discard it right after!) to make it presentable.

Allow me to go on a tangent and quickly check `dumpsys SurfaceFlinger`:

```
---------------------------------------------------------------------------------------------------------------------------------------------------------------
 rust.example.winit_android/android.app.NativeActivity#39518
  rel      0 |            1 |            0 |     DEVICE |          0 |    0    0 1096 2560 |    0.0    0.0 1096.0 2560.0 |                                              [*]
...
HWC2 display_id: 0
layer: 11369 name:                                                               VRI[NativeActivity]#0(BLAST Consumer)0 z: 0 composition: Device/Device alpha: 255 format:              RGBX_8888 dataspace:0x08810000 transform: 0/0/0 buffer_id: 0xb400007871ca9ab0 secure: 0
                 LayerBuffer:  width=[1152],  height=[2560],  unaligned_width=[1096],  unaligned_height=[2560]
                 LayerBufferFlags:  secure=[0],  interlace=[0],  hdr=[0]
                 ColorMetaData:  colorPrimaries=[1],  range=[1],  transfer=[1],  matrixCoefficients=[0](matrixCoefficients is not used for SDR layer)
...
|-----|---------------|-----------|------|-------------|--------------------------|---------------------|---------------------|----|------------|-----------|----|-----|----|
| Idx |   Comp Type   |   Split   | Pipe |    W x H    |          Format          |  Src Rect (L T R B) |  Dst Rect (L T R B) |  Z | Pipe Flags | Deci(HxV) | CS | Rng | Tr |
|-----|---------------|-----------|------|-------------|--------------------------|---------------------|---------------------|----|------------|-----------|----|-----|----|
|   0 |           SDE |    Pipe-1 |  139 | 1152 x 2560 |                RGBX_8888 |    0    0  548 2560 |    0    0  548 2560 |  0 | 0x00002001 |   0 x   0 |  1 |   1 |  1 |
|     |               |    Pipe-2 |  136 | 1152 x 2560 |                RGBX_8888 |  548    0 1096 2560 |  548    0 1096 2560 |  0 | 0x00002001 |   0 x   0 |  1 |   1 |  1 |
```

If you ignore the fact that my phone is a dualpipe (two layer mixers and two encoders, each getting half of the plane, to drive a 4k@120Hz screen), you can clearly see that the "composition type" of this `RGBX_8888` buffer of `1152x2560` pixels is "Snapdragon Display Engine" instead of some kind of software (or GPU) composition.

---

Completely separate from all this there's a more low-level API to [dispatch `HardwareBuffer`s (that we allocate ourselves!) to a surface](https://developer.android.com/ndk/reference/group/native-activity#asurfacetransaction_setbuffer) [^1]. Here we _could_ implement buffer aging correctly.

[^1]: Lots of caveats apply here:
      1. The API is still on a branch for the `ndk` crate, to be finished;
      2. NDK functions to retrieve a `SurfaceControl` for the root `NativeWindow` are broken right now and might only be fixed in Android 15 onwards;
      3. Requires API level 29.

notgull (approved):
LGTM for now, although we really should do something about pixel formats

PR #132, opened 2023-07-10
Use raw-window-handle version 0.6

notgull:
cc rust-windowing/raw-window-handle#125, rust-windowing/winit#2943

Testing out the new release before it becomes official.

### DIFF
--- a/src/wayland/mod.rs
+++ b/src/wayland/mod.rs
@@ -1,5 +1,11 @@
-use crate::{error::SwResultExt, util, Rect, SoftBufferError};
-use raw_window_handle::{WaylandDisplayHandle, WaylandWindowHandle};
+use crate::{
+    error::{InitError, SwResultExt},
+    util, Rect, SoftBufferError,
+};
+use raw_window_handle::{
+    HasDisplayHandle, HasRawDisplayHandle, HasRawWindowHandle, HasWindowHandle, RawDisplayHandle,
+    RawWindowHandle,
+};
 use std::{
     cell::RefCell,
     num::{NonZeroI32, NonZeroU32},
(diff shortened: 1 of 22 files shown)


ids1024:
We should probably add feature flags for raw-window-handle versions, like in https://github.com/rust-windowing/winit/pull/3126.

ids1024:
Actually, I guess we can't have a function that accepts either something implementing both `raw_window_handle_0_5::HasRawWindowHandle` and `raw_window_handle_0_6::HasRawWindowHandle`, because a type could implement both and there's no way to prefer one over the other...

Anyway, if it's going to support only 0.6, it should be released alongside the Winit version that adds support for 0.6.

ids1024:
CI seems to be failing because the benchmark fails to build. So that also needs to be updated.

src/wayland/mod.rs
```diff
@@ -17,17 +20,31 @@ use buffer::WaylandBuffer;
 
 struct State;
 
-pub struct WaylandDisplayImpl {
-    conn: Connection,
+pub struct WaylandDisplayImpl<D: ?Sized> {
+    conn: Option<Connection>,
```
ids1024, line 24: Instead of using an `Option`, I'd probably just rely on field order, and have a safety comment indicating this requirement.

There's also `ManuallyDrop`, though the docs for it seem to just recommend using field order for this sort of thing instead of dealing with it: https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html#manuallydrop-and-drop-order

src/wayland/mod.rs
```diff
@@ -17,17 +20,31 @@ use buffer::WaylandBuffer;
...
+    /// This has to be dropped *after* the `conn` field, because the `conn` field implicitly borrows
+    /// this.
+    _display: D,
 }
 
-impl WaylandDisplayImpl {
-    pub unsafe fn new(display_handle: WaylandDisplayHandle) -> Result<Self, SoftBufferError> {
-        // SAFETY: Ensured by user
-        let backend = unsafe { Backend::from_foreign_display(display_handle.display as *mut _) };
+impl<D: HasDisplayHandle + ?Sized> WaylandDisplayImpl<D> {
```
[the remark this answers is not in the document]
notgull: This way it allows for `Context<D>` to be coerced into `Context<dyn HasDisplayHandle>`, and be used like that.

src/wayland/mod.rs
notgull: I'd rather rely on `Option` than unsafe `ManuallyDrop` or fragile field ordering.

tests/present_and_fetch.rs
```diff
@@ -1,3 +1,5 @@
+// TODO: Once winit is updated again, restore this test.
```
ids1024, line 1: Winit is released, so presumably this can be updated. Though https://github.com/notgull/winit-test will need a release.

notgull: https://github.com/notgull/winit-test/pull/8

ids1024 (approved):

PR #134, opened 2023-07-13
Reintroduce Error trait impl for SoftBufferError

fenhl:
This trait implementation was removed in #116. It's not mentioned as a breaking change in [the changelog](https://github.com/rust-windowing/softbuffer/blob/master/CHANGELOG.md#030), so I'm guessing this was unintentional. This PR reintroduces the implementation.


notgull (approved):
Thanks!

PR #135, opened 2023-07-15
Add a DRM/KMS backend

### LINKED ISSUE #42: GBM/KMS platform
It may be useful to allow presenting CPU rendered buffers directly to the display on Linux.

raw-window-handle already supports GBM and DRM so it should be possible to use GBM or dumb buffers.

notgull:
This adds a DRM/KMS based backend to the system, as per #42. This system finds a CRTC and a connector, then uses that to create a frame buffer and a DUMB buffer that it can render to.

~~All of this is untested, hence the draft status.~~

I'm not going to be the maintainer for this backend, as I'm self-imposing a limit of two backends on myself. If anyone else with DRM/KMS experience or a motivating use case wants to maintain this backend, let me know. Otherwise I've marked it as unmaintained.

Closes #42

### DIFF
--- a/examples/drm.rs
+++ b/examples/drm.rs
@@ -10,12 +10,12 @@ mod imple {
 
     use std::num::NonZeroU32;
     use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd};
-    use std::path::{Path, PathBuf};
+    use std::path::Path;
     use std::time::{Duration, Instant};
 
     pub(super) fn entry() -> Result<(), Box<dyn std::error::Error>> {
         // Open a new device.
-        let device = Card::open()?;
+        let device = Card::find()?;
 
         // Create the softbuffer context.
         let context = unsafe {
@@ -164,54 +164,26 @@ mod imple {
     struct Card(std::fs::File);
 
     impl Card {
-        fn open() -> Result<Card, Box<dyn std::error::Error>> {
-            let mut enumerator = udev::Enumerator::new()?;
+        fn find() -> Result<Card, Box<dyn std::error::Error>> {
+            for i in 0..10 {
+                let path = format!("/dev/dri/card{i}");
+                let device = Card::open(path)?;
 
-            // Look for a drm system with name card[0-9]
-            enumerator.match_subsystem("drm")?;
-            enumerator.match_sysname("card[0-9]*")?;
+                // Only use it if it has connectors.
+                let handles = match device.resource_handles() {
+                    Ok(handles) => handles,
+                    Err(_) => continue,
+                };
+                
+                if !handles.connectors.is_empty() {
+                    return Ok(device);
+                }
+            }
 
-            // Get the first device.
-            let path = enumerator
-                .scan_devices()?
-                .filter(|device| {
-                    let seat_name = device
-                        .property_value("ID_SEAT")
-                        .map(|x| x.to_os_string())
-                        .unwrap_or_else(|| "seat0".into());
+            Err("No DRM device found".into())
+        }
 
-                    seat_name == "seat0"
-                })
-                .filter(|device| {
-                    let pci = match device.parent_with_subsystem(Path::new("pci")) {
-                        Ok(Some(pci)) => pci,
-                        _ => return false,
-                    };
-
-                    let id = match pci.attribute_value("boot_vga") {
-                        Some(id) => id,
-                        _ => return false,
-                    };
-
-                    id.to_str() == Some("1")
-                })
-                .find_map(|dev| dev.devnode().map(PathBuf::from))
-                .or_else(|| {
-                    enumerator
-                        .scan_devices()
-                        .expect("Failed to scan devices")
-                        .filter(|device| {
-                            let seat_name = device
-                                .property_value("ID_SEAT")
-                                .map(|x| x.to_os_string())
-                                .unwrap_or_else(|| "seat0".into());
-
-                            seat_name == "seat0"
-                        })
-                        .find_map(|dev| dev.devnode().map(PathBuf::from))
-                })
-                .ok_or("No drm device found")?;
-
+        fn open(path: impl AsRef<Path>) -> Result<Card, Box<dyn std::error::Error>> {
             let file = std::fs::OpenOptions::new()
                 .read(true)
                 .write(true)
--- a/examples/drm.rs
+++ b/examples/drm.rs
@@ -10,6 +10,7 @@ mod imple {
 
     use std::num::NonZeroU32;
     use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd};
+    use std::path::{Path, PathBuf};
     use std::time::{Duration, Instant};
 
     pub(super) fn entry() -> Result<(), Box<dyn std::error::Error>> {
@@ -119,6 +120,15 @@ mod imple {
             buffer.present()?;
 
             // Wait for the page flip to happen.
+            rustix::event::poll(
+                &mut [rustix::event::PollFd::new(
+                    &device,
+                    rustix::event::PollFlags::IN,
+                )],
+                -1,
+            )?;
+
+            // Receive the events.
             let events = device.receive_events()?;
             println!("Got some events...");
             for event in events {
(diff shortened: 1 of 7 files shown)


notgull:
I've tested this on my computer and it seems to work for my use case, so I think it's ready.

src/kms.rs
```diff
@@ -0,0 +1,329 @@
...
+            if buffer_width == width && buffer_height == height {
+                return Ok(());
+            }
+        }
+
+        // Create a new buffer.
+        let buffer = BufferSet::new(&self.display, width, height)?;
+
+        // Set the framebuffer in the CRTC info.
+        // TODO: This requires root access, find a way that doesn't!
```
tronical, line 180: I've had the same issue when implementing KMS support in Slint, and I've found using libseat to be a good solution. When set up, the daemon (for example logind or seatd) return you an FD that's has enough rights to set the frame buffer.

I think that doesn't affect the soft buffer API, but merely your example here (if you wanted to use it).

src/kms.rs
```diff
@@ -0,0 +1,329 @@
...
+                        .y
+                        .checked_add(rect.height.get())
+                        .and_then(|y| y.try_into().ok())
+                        .ok_or_else(err)?,
+                })
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        self.display
+            .dirty_framebuffer(self.fb, &rectangles)
```
tronical, line 282: I think this works but gives tearing. I suggest using two dumb buffers and flip between them. You could say that `present()` in soft buffer just posts it, but it's up to the application to wait for the page flip (although that's also tricky).

tronical (commented):
Awesome stuff, btw! Two small comments inside.

src/kms.rs
tronical: I think you just allocate two dumb buffers instead of one,map the one that you want to render into, create a frame buffer, and set it via set_crtc() when you're ready. In the next round you render into the second buffer, create a frame buffer, and then make it the new front buffer by calling `page_flip()`. 

As if you're using gbm, except that you're doing managing a back and front buffer yourself.

I don't know if the code we're using is 100% correct, but with gbm seems to work well like this: https://github.com/slint-ui/slint/blob/f8e9d6237999910db905b4ba84d2cd3520268d44/internal/backends/linuxkms/display/egldisplay.rs#L51

Without gbm I think you'd just allocate two dumb buffers (like you currently do in new()) and otherwise `present()` is pretty much the same. Perhaps some more magic in gbm's lock_buffer(), but I suspect that's more GL related (flushing pipeline?) than buffer related.

Take this with a grain of salt though ;-). I've been meaning to try to do this myself with Rust code ("double dumb buffers" with software rendering), but haven't gotten around it yet.

examples/drm.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    impl Device for Card {}
+    impl CtrlDevice for Card {}
+}
+
+#[cfg(not(kms_platform))]
+mod imple {
+    pub(super) fn entry() -> Result<(), Box<dyn std::error::Error>> {
+        eprintln!("This example requires the `kms` feature.");
```
ids1024, line 165: The normal way to do this would be to define an `[[example]]` in `Cargo.toml` with `required-features`.

src/kms.rs
ids1024: Yes, the `libseat` crate is your best option here to make this work without root wherever possible.

src/kms.rs
```diff
@@ -0,0 +1,329 @@
...
+            )
+            .ok();
+    }
+}
+
+impl BufferImpl<'_> {
+    #[inline]
+    pub fn pixels(&self) -> &[u32] {
+        // drm-rs doesn't let us have the immutable reference... so just use a bunch of zeroes.
+        // TODO: There has to be a better way of doing this!
```
ids1024, line 244: Well, the `BorrowStack` hack would work here.

But it looks like `DumbMapping` is already just a wrapper around a `&'a mut [u8]`, so could just open a PR for drm-rs to implement `AsRef<[u8]>` in addition to the `AsMut<[u8]>` implementation.

src/kms.rs
```diff
@@ -0,0 +1,329 @@
...
+                    .copied()
+                    .swbuf_err("failed to find a primary CRTC")?
+            }
+        };
+
+        // Use a preferred connector or just select the first one.
+        let connector = handles
+            .connectors
+            .iter()
+            .flat_map(|handle| display.get_connector(*handle, false))
```
ids1024, line 141: Hm, it seems strange for softbuffer to be dealing with identifying what connector to use. Should that not be part of the `DrmRawWindowHandle`?

notgull: It should, I'll file that PR. But it probably won't be released until rwh v0.6, so I think this solution is best for now.

notgull:
drm_example

Tested on my laptop

src/kms.rs
notgull: Actually nevermind, I found a good solution for this for now.

examples/drm.rs
notgull: As this example is also platform specific, I don't think that works here.

ids1024:
`page_flip` only queues a page flip. If an application were presenting buffers repeatedly without waiting, I believe it could overwrite a buffer while/before it appears on screen.

So I believe like on Wayland, `buffer_mut` should wait for buffer release. I guess it would have to `poll` the device fd, and use `drm::control::device::receive_events` to see if it's gotten `Event::PageFlip` yet.

...but would Winit or another "windowing" backend using DRM also poll the same fd? Is there any way to manage this?

examples/drm.rs
ids1024: Ah right, not sure about a better way to deal with that.

examples/drm.rs
```diff
@@ -0,0 +1,186 @@
...
+        while Instant::now().duration_since(start) < Duration::from_secs(2) {
+            tick += 1;
+            println!("Drawing tick {tick}");
+
+            // Start drawing.
+            let mut buffer = surface.buffer_mut()?;
+            draw_to_buffer(&mut buffer, tick);
+            buffer.present()?;
+
+            // Wait for the page flip to happen.
```
ids1024, line 121: I don't think this actually waits on anything currently. To "wait for the page flip" it would have to `poll` the fd and read any events when it's readable until `Event::PageFlip` occurs, right?

ids1024:
The example currently fails on the laptop I'm using since it has an Nvidia card, which for some reason is `card0`, while the only connected connector is on the iGPU, `card1`.

There isn't a good way to determine the right output, and the display probably shouldn't try to deal with setting all connected outputs, but maybe it should iterate over cards and chose the first one it finds a connector on?

notgull:
> `page_flip` only queues a page flip. If an application were presenting buffers repeatedly without waiting, I believe it could overwrite a buffer while/before it appears on screen.
> 
> 
> 
> So I believe like on Wayland, `buffer_mut` should wait for buffer release. I guess it would have to `poll` the device fd, and use `drm::control::device::receive_events` to see if it's gotten `Event::PageFlip` yet.
> 
> 
> 
> ...but would Winit or another "windowing" backend using DRM also poll the same fd? Is there any way to manage this?

I'm not sure if there is a way to do this on DRM. Please tell me if I'm wrong, but I think the current way of doing things is the only way.

ids1024:
I guess `buffer_mut` could have a comment noting it blocks for buffer release on Wayland, and on DRM, should not be called until after waiting for pageflip?

Looking at https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/egl/drivers/dri2/platform_drm.c and https://github.com/libsdl-org/SDL/blob/main/src/video/kmsdrm/SDL_kmsdrmopengles.c, it seems like EGL manages this in some way with functions called `gbm_surface_lock_front_buffer` and `gbm_surface_lock_front_buffer`, but I don't entirely follow the logic there and it's not clear that anything like that could make sense in softbuffer. I guess that's something Glutin would have to deal with for DRM support.

notgull:
> I guess `buffer_mut` could have a comment noting it blocks for buffer release on Wayland, and on DRM, should not be called until after waiting for pageflip?

Done.

I was mostly following https://github.com/rust-windowing/glutin/pull/1403, which appears to use fencing in a way that I don't know how to accommodate in `softbuffer` with DUMB buffers. As far as I know there's not a real way to say "fence a DUMB buffer off" like there is in GBM.

examples/drm.rs
```diff
@@ -0,0 +1,215 @@
...
+    struct Card(std::fs::File);
+
+    impl Card {
+        fn find() -> Result<Card, Box<dyn std::error::Error>> {
+            for i in 0..10 {
+                let path = format!("/dev/dri/card{i}");
+                let device = Card::open(path)?;
+
+                // Only use it if it has connectors.
+                let handles = match device.resource_handles() {
```
ids1024, line 173: This test isn't really sufficient for a multi-gpu system since multiple GPUs will have DRM "connectors" even if they aren't physically exposed in any way.

It should check for one that's currently connected.

ids1024:
Seems like `DIRTYFB` is returning `ENOSYS`. Removing that call makes the example work.

I guess that's expected with some drivers? Don't know if simply ignoring `ENOSYS` and doing nothing else is correct here. DRM isn't the best documented thing in the world...

ids1024 (approved):

Issue #124, opened 2023-06-12
Segmentation fault on macOS if resize() is not called


kpreid:
Write a program that neglects to call `Surface::resize()`, for example by deleting all the size and rendering lines in `examples/rectangle.rs`:

```rust
Event::RedrawRequested(window_id) if window_id == window.id() => {
    let mut buffer = surface.buffer_mut().unwrap();
    buffer.present().unwrap();
}
```

Run this example on macOS, and observe the crash log — it looks like softbuffer gave a [`dangling`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.dangling) pointer as the buffer pointer, by not checking whether the `BufferImpl::buffer` had nonzero capacity.

```text
...
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000004
Exception Codes:       0x0000000000000001, 0x0000000000000004
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [58050]

VM Region Info: 0x4 is not in any region.  Bytes before following region: 140737486381052
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      VM_ALLOCATE              7fffffe1e000-7fffffe1f000 [    4K] r-x/r-x SM=ALI  

Thread 0 Crashed:: main Dispatch queue: com.apple.main-thread
0   CoreGraphics                  	    0x7ff806a2cfb7 ERROR_CGDataProviderCreateWithData_BufferIsNotReadable + 39
1   CoreGraphics                  	    0x7ff806bb7096 check_clients_buffer + 118
2   CoreGraphics                  	    0x7ff806a2cf66 CGDataProviderCreateWithData + 35
3   rectangle                     	       0x10cd2ea5e core_graphics::data_provider::CGDataProvider::from_buffer::hf5a7344b1dce02de + 222
4   rectangle                     	       0x10cd2fce8 softbuffer::cg::BufferImpl::present::h33f92d08748cdedf + 88
5   rectangle                     	       0x10cd348b7 softbuffer::BufferDispatch::present::hf5096c1071865b18 + 55
6   rectangle                     	       0x10cd34417 softbuffer::Buffer::present::heb646d1ee9f8107d + 55
[...]

PR #137, opened 2023-07-25
cg: Require non-zero width/height; panic instead of segfault

ids1024:
Fixes https://github.com/rust-windowing/softbuffer/issues/124.


notgull (approved):
Looks good to me!

The macOS maintainer is AFK at the moment, by the way

PR #140, opened 2023-08-17
orbital: Use `syscall::MAP_SHARED` for mapping

ids1024:
This seems to be required to work correctly since https://gitlab.redox-os.org/redox-os/orbital/-/merge_requests/51.

Orbclient had a similar change: https://gitlab.redox-os.org/redox-os/orbclient/-/commit/9097e71c9a61f0f654113881a485e16ce3854c62.


jackpot51 (approved):

jackpot51:
Yes, that is correct.

PR #141, opened 2023-08-27
v0.3.1

ids1024:
Softbuffer is broken on recent builds of Redox without the fix here, so it's helpful to have a new release that fixes it. We seem to have accumulated a few non-breaking fixes, so it seems like a good enough time to do a release anyway.

Disabling `fetch` on Windows could be considered breaking, but it doesn't change the API, was broken anyway, and is likely unused. So I don't think that's an issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,14 @@
-# UNRELEASED
+# 0.3.1
 
 * On X11, fix the length of the returned buffer when using the wire-transferred buffer.
 * On Web, fix incorrect starting coordinates when handling buffer damage.
+* Or Redox, use `MAP_SHARED`; fixing behavior with latest Orbital.
+* Error instead of segfault on macOS if size isn't set.
+* Add `OffscreenCanvas` support in web backend.
+* Add DRM/KMS backend, for running on tty without X/Wayland.
+* Make `fetch` error on Windows, where it wasn't working correctly.
+* Implement `Error` trait for `SoftBufferError`.
+* Dependency updates.
 
 # 0.3.0
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "softbuffer"
-version = "0.3.0"
+version = "0.3.1"
 edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "Cross-platform software buffer"


ids1024:
I have no idea what's going on with the link errors for the Windows CI; I guess some testing there is warranted before a release, though I'm not currently booted into Windows...

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# 0.3.1
+
+* DRM/KMS backend
+* Use `MAP_SHARED` on Redox; fixing behavior with latest Orbital
+* `OffscreenCanvas` support in web backend
+* Error instead of segfault on macOS if size isn't set
+* Make `fetch` error on Windows, where it wasn't working correctly
+* Implement `Error` trait for `SoftBufferError`
+* Dependency updates
```
notgull, line 9: I think that some of these lines are useful as commit messages for us as developers, but as less useful as changes for users. For instance:

```
* DRM/KMS backend
```

Should probably be something more like:

```
* Add support for using `softbuffer` on raw TTYs, using the DRM/KMS interface.
```

Also, it would be nice if these entries were accompanied by PR numbers where applicable.

notgull (changes requested):

notgull:
I'll look into the Windows errors

jackpot51:
Any updates on this?

ids1024:
It looks like https://github.com/rust-windowing/softbuffer/issues/142 isn't really related to softbuffer, so that probably doesn't block a release...

May be better to release before merging https://github.com/rust-windowing/softbuffer/pull/149, since winit will still be using 0.30 (pending https://github.com/Smithay/client-toolkit/pull/403, etc.)

jackpot51:
I agree. @notgull what do you think, can we do a release if the windows issue is not a regression?

ids1024:
https://github.com/rust-windowing/softbuffer/pull/150 could be good to get in the release, if @daxpedda thinks that PR is good. (The Web backend is probably the one I'm least familiar with.)

daxpedda:
> #150 could be good to get in the release, if @daxpedda thinks that PR is good. (The Web backend is probably the one I'm least familiar with.)

Indeed, would be nice to wait for that, I'm reviewing it as we speak.

notgull:
> I agree. @notgull what do you think, can we do a release if the windows issue is not a regression?

I'm fine with it.

notgull (approved):

jackpot51:
@ids1024 are you able to tag/publish a release?

ids1024:
I already released on crates.io. I hadn't thought to tag a release on Github, but I've done that now.

PR #148, opened 2023-09-06
Fix length of returned buffer on x11 when using the wire-transferred buffer

tronical:
total_len() computes the number of bytes, while the wire Vec holds u32.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# UNRELEASED
+
+* On X11, fix the length of the returned buffer when using the wire-transferred buffer.
+
 # 0.3.0
 
 * On MacOS, the contents scale is updated when set_buffer() is called, to adapt when the window is on a new screen (#68).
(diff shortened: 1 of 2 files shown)


tronical:
This is reproducible using this simple "test case"

```patch
diff --git a/examples/winit.rs b/examples/winit.rs
index fe5a82b..2a7578c 100644
--- a/examples/winit.rs
+++ b/examples/winit.rs
@@ -42,6 +42,7 @@ fn main() {
                     .unwrap();

+                assert_eq!(buffer.len(), width as usize * height as usize);
                 for y in 0..height {
                     for x in 0..width {
                         let red = x % 255;
```

notgull (approved):
Thanks!

tronical:
Would you folks be willing to release a 0.3.1 with this fix included?

The overall use-case we noticed is that when using softbuffer under WSL with X11, there's no shm transport and thus the wire transfer is chosen, which is unfortunately broken. It's quite easy to run into this though by simply choosing a distro that's not using wayland, for example one with XFCE.

PR #149, opened 2023-09-08
wayland: Update to wayland-client 0.31

ids1024:


.github/workflows/ci.yml
```diff
@@ -77,9 +77,13 @@ jobs:
         targets: ${{ matrix.platform.target }}
         components: clippy, rust-src
 
+    - name: Install libwayland
+      if: (matrix.platform.os == 'ubuntu-latest')
+      run: sudo apt-get update && sudo apt-get install libwayland-dev
+
     - name: Install GCC Multilib
       if: (matrix.platform.os == 'ubuntu-latest') && contains(matrix.platform.target, 'i686')
-      run: sudo apt-get update && sudo apt-get install gcc-multilib
```
notgull, line 82: Huh?

notgull (approved):
Looks good to me aside from one minor conclusion.

.github/workflows/ci.yml
ids1024: The way Github chose to display the diff is interesting, but this line now runs just `apt-get install` and not `apt-get update`, since the previous step will already have done an `apt-get update`.

notgull: Alright, that makes sense to me

PR #150, opened 2023-09-08
fix(web): optimize and fix present_with_damage

CBenoit:
This patch fixes the `present_with_damage` implementation for the web backend. `CanvasRenderingContext2D`’s `putImageData` was used incorrectly.

As per [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/putImageData):

> imageData
>
>   An ImageData object containing the array of pixel values.
>
> dx
>
>   Horizontal position (x coordinate) at which to place the image data
>   in the destination canvas.
>
> dy
>
>   Vertical position (y coordinate) at which to place the image data
>   in the destination canvas.
>
> dirtyX
>
>   Horizontal position (x coordinate) of the top-left corner from which
>   the image data will be extracted. Defaults to 0.
>
> dirtyY
>
>   Vertical position (y coordinate) of the top-left corner from which
>   the image data will be extracted. Defaults to 0.

As such, `dx` is different from `dirtyX` and `dy` is different from `dirtyY`.

When the actual `ImageData` is as big as the canvas itself, `dx` and `dy` should always be `0.0` and `0.0`, and only `dirtyX` and `dirtyY` are actually representing the top-left coordinates for the dirty region. The dirty region is relative to the image data.

This patch also optimizes the BGRX to RGBA conversion by only creating a bitmap for the smallest region enclosing all the damaged regions. This does not affect the behavior of `present`.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,6 +1,10 @@
 // Not needed on all platforms
 #![allow(dead_code)]
 
+use std::cmp;
+use std::num::NonZeroU32;
+
+use crate::Rect;
 use crate::SoftBufferError;
 
 /// Takes a mutable reference to a container and a function deriving a
@@ -42,6 +46,41 @@ impl<'a, T: 'static + ?Sized, U: 'static + ?Sized> BorrowStack<'a, T, U> {
         // mutable reference is unique.
         unsafe { &mut *self.container }
     }
+}
+
+/// Calculates the smallest `Rect` necessary to represent all damaged `Rect`s.
+pub(crate) fn union_damage(damage: &[Rect]) -> Option<Rect> {
+    struct Region {
+        left: u32,
+        top: u32,
+        bottom: u32,
+        right: u32,
+    }
+
+    let region = damage
+        .iter()
+        .map(|rect| Region {
+            left: rect.x,
+            top: rect.y,
+            right: rect.x + rect.width.get(),
+            bottom: rect.y + rect.height.get(),
+        })
+        .reduce(|mut prev, next| {
+            prev.left = cmp::min(prev.left, next.left);
+            prev.top = cmp::min(prev.top, next.top);
+            prev.right = cmp::max(prev.right, next.right);
+            prev.bottom = cmp::max(prev.bottom, next.bottom);
+            prev
+        })?;
+
+    Some(Rect {
+        x: region.left,
+        y: region.top,
+        width: NonZeroU32::new(region.right - region.left)
+            .expect("`right` must always be bigger then `left`"),
+        height: NonZeroU32::new(region.bottom - region.top)
+            .expect("`bottom` must always be bigger then `top`"),
+    })
 }
 
 #[cfg(test)]
--- a/src/web.rs
+++ b/src/web.rs
@@ -146,7 +146,12 @@ impl WebImpl {
 
         let mut damage_iter = damage.iter();
 
-        let first_rect = damage_iter.next().expect("at least one damage rectangle");
+        let first_rect = if let Some(rect) = damage_iter.next() {
+            rect
+        } else {
+            // If there is no damage, there is nothing to do
+            return Ok(());
+        };
 
         struct UnionRegion {
             top: u32,
--- a/src/web.rs
+++ b/src/web.rs
@@ -2,6 +2,11 @@
 
 #![allow(clippy::uninlined_format_args)]
 
+use std::cmp;
+use std::convert::TryInto;
+use std::marker::PhantomData;
+use std::num::NonZeroU32;
+
 use js_sys::Object;
 use raw_window_handle::WebWindowHandle;
 use wasm_bindgen::{JsCast, JsValue};
@@ -11,9 +16,6 @@ use web_sys::{OffscreenCanvas, OffscreenCanvasRenderingContext2d};
 
 use crate::error::SwResultExt;
 use crate::{Rect, SoftBufferError};
-use std::convert::TryInto;
-use std::marker::PhantomData;
-use std::num::NonZeroU32;
 
 /// Display implementation for the web platform.
 ///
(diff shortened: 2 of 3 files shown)


src/web.rs
```diff
@@ -181,8 +225,10 @@ impl WebImpl {
             self.canvas
                 .put_image_data(
                     &image_data,
-                    rect.x.into(),
-                    rect.y.into(),
+                    union_region.left.into(),
+                    union_region.top.into(),
+                    (rect.x - union_region.left).into(),
+                    (rect.y - union_region.top).into(),
```
CBenoit, line 231: > When the actual `ImageData` is as big as the canvas itself, `dx` and `dy` should always be `0.0` and `0.0`, and only `dirtyX` and `dirtyY` are actually representing the top-left coordinates for the dirty region. The dirty region is relative to the image data.

This is what I’m referring to here. Without the optimization we have `union_region.left = 0` and `union_region.top = 0`, resulting in something like:
```rust
                    0.0,
                    0.0,
                    rect.x.into(),
                    rect.y.into(),
```

CBenoit:
I addressed the clippy lint.

src/web.rs
```diff
@@ -138,17 +140,57 @@ impl WebImpl {
...
 
     fn present_with_damage(&mut self, damage: &[Rect]) -> Result<(), SoftBufferError> {
-        let (width, _height) = self
+        let (buffer_width, _buffer_height) = self
             .size
             .expect("Must set size of surface before calling `present_with_damage()`");
+
+        let mut damage_iter = damage.iter();
+
+        let first_rect = damage_iter.next().expect("at least one damage rectangle");
```
CBenoit, line 149: On second thought, maybe this should not panic, because it’s actually possible for someone to write this code:
```rust
surface.present_with_damage(&[])?;
```

daxpedda: Sorry for the force-push here, just fixed the compile error.

src/web.rs
daxpedda: I'm not following the logic here, why do we pass `union_region.left` to `dx`?

As far as I understand `dx` is where to draw relative to the destination image, `union_region.left` is the X coordinate of our coalesced dirty `Rect`s compared to the destination image.

daxpedda: I assume it's what you were trying to explain in OP, but I don't seem to really get it. As far as I can tell the original implementation worked correctly.

CBenoit: I’m sorry, my explanation was likely not clear enough.

> As far as I understand `dx` is where to draw relative to the destination image

Yes, to be pedantic, it is where to draw the _image data_ on the _destination canvas_.

The original implementation is [building a bitmap / image data for the _whole canvas_](https://github.com/rust-windowing/softbuffer/blob/9b44f2eaea136cae30543ecbad96827af7ff579c/src/web.rs#L144-L150).
As such, `dx` (respectively `dy`) was supposed to be `0.0`, because in that case the top left pixel of the image data is matching the top left pixel in the destination canvas.

However, the original implementation is [not passing `0.0`](https://github.com/rust-windowing/softbuffer/blob/9b44f2eaea136cae30543ecbad96827af7ff579c/src/web.rs#L182-L188), it passes [`rect.x` (respectively `rect.y`)](https://github.com/rust-windowing/softbuffer/blob/9b44f2eaea136cae30543ecbad96827af7ff579c/src/web.rs#L285-L293).

Next, `dirtyX` is the X-cordinate of the top-left corner of the damaged region _inside the image data_, NOT in the destination canvas.

To clarify: this does work correctly when the damage rectangle have _exactly_ the same dimension as the canvas, i.e.: `present` is not broken for this reason because [it happens to use the only valid damage rectangle](https://github.com/rust-windowing/softbuffer/blob/9b44f2eaea136cae30543ecbad96827af7ff579c/src/web.rs#L325-L330) for which the computation gives the correct result.

Here is an example. Let’s assume we have a 10x10 pixels canvas and let `data` be the image data for the whole canvas (again, 10x10) and `dirty = { top: 5, left: 5, bottom: 8, right: 8 }` be the damaged region.

Given the above points, the correct call is:
```js
canvas.putImageData(data, 0, 0, 5, 5, 3, 3);
```
Which roughly means "write the 10x10 image data at (0, 0) in the canvas with a dirty region from (5, 5) to (8, 8) relative to the image data". In this case only the pixels from (5, 5) to (8, 8) (the dirty region) are extracted from the image data and actually drawn.

The original implementation is making this call:
```js
canvas.putImageData(data, 5, 5, 5, 5, 3, 3);
```

I hope it helps clarifying a bit.

src/web.rs
```diff
@@ -138,17 +140,62 @@ impl WebImpl {
...
+            right: (first_rect.x + first_rect.width.get()),
+        };
+
+        let union_region = damage_iter.fold(union_region, |mut union, rect| {
+            union.top = cmp::min(union.top, rect.y);
+            union.left = cmp::min(union.left, rect.x);
+            union.bottom = cmp::max(union.bottom, rect.y + rect.height.get());
+            union.right = cmp::max(union.right, rect.x + rect.width.get());
+            union
+        });
```
notgull, line 176: You should use `reduce` instead of `fold` here.

CBenoit: I agree this is almost perfect for `reduce`, but the accumulator is not a `Rect` so I can’t exactly do that

src/web.rs
CBenoit: > why do we pass `union_region.left` to `dx`?

Building upon the previous answer: in my patch, the image data holds the pixels for the union of all the damaged regions only, and as such does not necessarily covers the left-most part of the canvas as the original implementation does.

daxpedda: I honestly can't interpret it this way from the MDN documentation, but the WHATWG specification was much clearer to me and seems to be in agreement with what you are saying. I also tested this locally and can confirm that this is correct.

Thanks!

src/web.rs
daxpedda: I used a `Iterator::map()` to do the trick.

src/util.rs
```diff
@@ -44,6 +48,41 @@ impl<'a, T: 'static + ?Sized, U: 'static + ?Sized> BorrowStack<'a, T, U> {
...
+
+    Some(Rect {
+        x: region.left,
+        y: region.top,
+        width: NonZeroU32::new(region.right - region.left)
+            .expect("`right` must always be bigger then `left`"),
+        height: NonZeroU32::new(region.bottom - region.top)
+            .expect("`bottom` must always be bigger then `top`"),
+    })
+}
```
daxpedda, line 84: I moved this to `util` as I think this might become useful when implementing #98 for other backends as well.

daxpedda (approved):
Thanks @CBenoit!

Apart from the fix this is also a great performance improvement!

daxpedda:
Just waiting for somebody to take a look at https://github.com/rust-windowing/softbuffer/pull/150#discussion_r1320521809 before I merge this.

@notgull, would be appreciated.

src/web.rs
CBenoit: Nice, this is actually much better

src/web.rs
CBenoit: To be fair, it tookme some time playing around with the visual artifacts to finally reach this conclusion; the MDN doc could be clearer

CBenoit:
Thank you for the prompt review!

src/util.rs
notgull: I'm fine with this

notgull (approved):
Looks good to me!

PR #163, opened 2023-10-13
Update `drm`, `nix`, and `criterion` dependencies

ids1024:


ids1024:
It seems `clap`, used by `criterion`, wants Rust 1.70.0.

notgull:
Let's pin the version of `criterion` for now, it's only a dev dependency.

notgull (approved):
Thanks!

Issue #147, opened 2023-09-04
Use rustix instead of nix


notgull:
This would allow us to use safer, direct system calls instead of `libc`.

Blocked on https://github.com/bytecodealliance/rustix/issues/797

ids1024:
With https://github.com/bytecodealliance/rustix/pull/848 and https://github.com/Smithay/wayland-rs/pull/658, we should be able to eliminate nix in the Wayland backend and its dependencies.

But yeah, there's still SYSV shm. Such an annoying API...

notgull:
It may be possible to rewrite the X backend to use POSIX shared memory as well as [`attach_fd`](https://docs.rs/x11rb/latest/x11rb/protocol/shm/fn.attach_fd.html) to get around this issue.

ids1024:
Yeah. That's an extension added by Xorg that may not be available everywhere, but I'm not sure it's likely to be an issue anywhere softbuffer would be used. This doesn't apply with network transparency anyway, and you're not likely to be running anything other than Xorg otherwise, or an extremely old version of Xorg.

ids1024:
Actually `shm*` are imported from `nix::libc`, so we could just depend on `libc` directly to eliminate `nix` there. It could be added to `rustix` later for a safe API and support for direct syscalls on Linux.

Though like Smithay, the kms backend here uses ` nix::errno::Errno` to compare errors from the `drm` crate. Which is a bit awkward to deal with.

ids1024:
I think this can be considered resolved now.

PR #164, opened 2023-10-13
Use `rustix`/`libc` instead of `nix`

ids1024:
Partly addresses https://github.com/rust-windowing/softbuffer/issues/147, though it would still be desirable to have a good safe API for SYSV shm in Rustix. But using `libc` directly for now is no worse than using the `nix::libc` re-export, so we don't lose anything.


notgull (approved):
Thanks!

PR #165, opened 2023-10-14
Use POSIX shared memory in X11 backend

notgull:
The X11 backend used System-V shared memory to communicate shared buffers to the server, but rustix does not support SysV SHM so we had to use libc in this backend. However, there is an alternate X11 API which uses POSIX shared memory, which rustix does support. This PR switches the X11 backend to POSIX shared memory and purges the previous usages of libc.

The goal is to remove libc totally. Therefore this PR also removes the default libc dependency from rustix.

### DIFF
--- a/src/x11.rs
+++ b/src/x11.rs
@@ -7,12 +7,14 @@
 
 use crate::error::SwResultExt;
 use crate::{Rect, SoftBufferError};
-use libc::{shmat, shmctl, shmdt, shmget, IPC_PRIVATE, IPC_RMID};
 use raw_window_handle::{XcbDisplayHandle, XcbWindowHandle, XlibDisplayHandle, XlibWindowHandle};
-use std::ptr::{null_mut, NonNull};
+use rustix::{fd, mm, shm as posix_shm};
 use std::{
-    fmt, io, mem,
+    fmt,
+    fs::File,
+    io, mem,
     num::{NonZeroU16, NonZeroU32},
+    ptr::{null_mut, NonNull},
     rc::Rc,
     slice,
 };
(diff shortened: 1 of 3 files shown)


src/x11.rs
```diff
@@ -758,6 +767,45 @@ impl Drop for X11Impl {
     }
 }
 
+/// Create a shared memory identifier.
+fn create_shm_id() -> io::Result<fd::OwnedFd> {
```
ids1024, line 771: We should move the `create_memfile` logic in the `wayland` backend to somewhere it can be used in both backends, instead of duplicating things.

Using `memfd_create` is better than `shm_open` where supported.

ids1024:
> The goal is to remove libc totally. 

Personally I don't see that much value to completely avoiding the `libc` dependency as long as the crate requires `std` and there isn't a version of std available that doesn't link to the C standard library. But I don't specifically have any issue with the changes in question.

src/x11.rs
notgull: I'm not sure if memfiles are supported for X11 MIT-SHM

ids1024: I can't really find any documentation for `ShmAttachFd` beyond:

> Create a shared memory segment.  The file descriptor will be mapped at offset
zero, and the size will be obtained using fstat().  A zero size will result in a
Value error.

> The file descriptor the server should mmap().

So any fd it can `fstat` to get the size and `mmap` should be fine.

https://lists.freedesktop.org/archives/xorg-devel/2013-October/038490.html

> I was hoping for a patch to shmproto.txt, but there isn't one, le sigh.
We should really fix the MIT-SHM docs, Jon's reverse-engineered spec pdf
is full of enough handwave to be embarassing for us.

Lol. I guess the fact https://gitlab.freedesktop.org/xorg/proto/xcbproto/-/blob/master/src/shm.xml exists means the blocker there is addressed, but no one has added documentation for version 1.2 of the protocol yet.

notgull:
> Personally I don't see that much value to completely avoiding the libc dependency as long as the crate requires std and there isn't a version of std available that doesn't link to the C standard library. But I don't specifically have any issue with the changes in question.

I think there's progress on this front.

PR #169, opened 2023-10-29
Fix missing paragraph break in the docs for `Buffer`

Liamolucko:
Right now the bit of the docs for `Buffer` discussing which platforms require copying looks like this:

Screenshot 2023-10-29 at 5 46 01 pm

'Currently `Buffer::present` must block copying image data on:' is meant to be a new heading for the last two dot-points, but currently gets interpreted as part of the dot-point for Orbital since there's no line break before it. This PR fixes that, making it look like it's supposed to:

Screenshot 2023-10-29 at 5 45 33 pm


notgull (approved):
Thanks!

PR #171, opened 2023-11-01
Enable creating an X11 display without connection

notgull:
For pure-Rust connections, there is no XCB connection that can be passed into the display constructor. Thus, this PR enables the display to be created without needing an XCB or Xlib handle, by providing `None` in the constructor.

cc rust-windowing/raw-window-handle#120

### DIFF
--- a/examples/libxcb.rs
+++ b/examples/libxcb.rs
@@ -6,7 +6,7 @@ mod example {
         DisplayHandle, RawDisplayHandle, RawWindowHandle, WindowHandle, XcbDisplayHandle,
         XcbWindowHandle,
     };
-    use std::{num::NonZeroU32, ptr::NonNull};
+    use std::{env, num::NonZeroU32, ptr::NonNull};
     use x11rb::{
         connection::Connection,
         protocol::{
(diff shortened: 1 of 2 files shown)


ids1024:
So if this were used along with winit (with a version of winit that uses x11rb without the xcb feature), there would be two sockets to the X server? Theoretically that seems unnecessary, but isn't really a problem and should work fine with X. That is probably better than exposing the version of x11rb in the public API.

Something like this wouldn't be possible with Wayland though.

The implementation is fairly straightforward, but it would be good to have some way to test this. And this still depends on xcb? I guess if the goal is to accommodate pure-Rust connections, we should use pure x11rb here, and have xcb dependencies/features behind a default feature flag?

notgull:
> The implementation is fairly straightforward, but it would be good to have some way to test this. 

Yes, I'll add a test case.

> And this still depends on xcb? I guess if the goal is to accommodate pure-Rust connections, we should use pure x11rb here, and have xcb dependencies/features behind a default feature flag?

Probably best for another PR, for bisection purposes.

ids1024 (approved):

PR #173, opened 2023-11-06
examples: Trigger event-loop exit on `Escape` key

MarijnS95:
In visual examples it is convenient to just press the escape key to exit the demo.


notgull (approved):
Thanks!

PR #178, opened 2023-11-14
Update `drm` to 0.11

ids1024:
Drm-rs now no longer requires separate generated bindings for each platform, so this fixes the build on architectures/OSes where drm didn't have bindings.

It also now uses Rustix instead of Nix, but Nix is still used by the Wayland and X11 backends, until the next wayland-rs and x11rb release.

This is not a breaking change since `drm` isn't exposed in the API.


notgull (approved):
Looks good to me!

PR #179, opened 2023-11-14
Update `winit_test` and re-enable `tests/present_and_fetch`

ids1024:


ids1024:
Looks like the test needs to be updated for WASM. The `SIGSEGV` on macOS is more concerning though.

madsmtm:
Hmm, yeah, that _is_ troubling, I'm looking into it.

madsmtm:
Oh, I thought I had migrated `softbuffer` to use `objc2` already, it would've prevented this issue...

The issue is fairly simple once you see it, both `CALayer` and the `window` reference are not retained (in part due to https://github.com/servo/core-foundation-rs/issues/643), so I think the window ends up being double-freed (or free'd before the CALayer has a chance to release it's own reference to it).

daxpedda:
This is unfortunately not going to work on Wasm, because `Window::inner_size()` will always return `None` on `Window` creation. You have to wait for a `Resized` event for that to work properly.

I guess this is an issue in `winit-test`, see https://github.com/notgull/winit-test/issues/11.

tests/present_and_fetch.rs
```diff
@@ -1,29 +1,20 @@
...
+        use winit::platform::web::WindowBuilderExtWebSys;
+
+        builder = builder.with_append(true);
     }
 
+    let window = builder.build(elwt).unwrap();
+
     // winit does not wait for the window to be mapped... sigh
     #[cfg(not(target_arch = "wasm32"))]
     std::thread::sleep(std::time::Duration::from_millis(1));
```
daxpedda, line 20: Just leaving this here as another argument for https://github.com/notgull/winit-test/issues/11: my system needed longer to work correctly.

notgull:
The test was removed as part of #214, so I'll close this PR. We will revisit this in the future.

PR #186, opened 2023-12-13
WIP Format API

ids1024:
Very initial implementation for https://github.com/rust-windowing/softbuffer/issues/98. No conversion support, and Wayland only, but allows listing available formats and selecting which to use. Seems good to get a start on this and see how things work.

It should be easy to add basic format conversion support swapping the first and third component. I'm not sure exactly about alpha. If a platform always uses alpha, a format like `BGRX` can't be supported without conversion (setting the fourth component to 255). If a platform doesn't support alpha, formats with alpha probably shouldn't be supported even with conversion? Instead of just listing `Format`, it could list structs with that and a `native: bool` field to indicate that the format is native and doesn't require conversion...

### DIFF
--- a/examples/winit.rs
+++ b/examples/winit.rs
@@ -24,7 +24,8 @@ fn main() {
     }
 
     let context = softbuffer::Context::new(window.clone()).unwrap();
-    let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+    let mut surface =
+        softbuffer::Surface::new(&context, window.clone(), softbuffer::Format::BGRX).unwrap();
 
     event_loop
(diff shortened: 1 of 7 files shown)


ids1024:
Handling conversion isn't hard (unless we add more unusual formats), though I'm thinking about how I'd want to integrate it.
* A conversion API that is internally used within each backend and called into by the backend's implementation of (at least) `pixels`/`pixels_mut`/`present`
* Code in toplevel `pixels`/`present`/etc. that either calls to the conversion code, or the backend directly
* A new variant of `BufferDispatch` that does conversion first, then calls through to the backend.

PR #192, opened 2024-01-10
0.3 backports

ids1024:
A new 0.3 release is useful for anything still relying on raw-window-handle 0.5. Given it isn't too hard. In particular https://github.com/iced-rs/iced/pull/2190.

Is https://github.com/rust-windowing/softbuffer/pull/180 also needed here? Or was that issue introduced later? Easy enough to add.

Leaving as a draft until I test on macOS (and make sure nothing has buffer age bugs now?). And we should also do a 0.4.x release with the current fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.3.4
+* Fix buffer age on Wayland. (#191)
+* Update `drm` to 0.11. (#178)
+  * Fixes build on architectures where drm-rs did not have generated bindings.
+
 # 0.3.3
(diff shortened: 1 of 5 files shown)


ids1024:
Okay, I think https://github.com/rust-windowing/softbuffer/pull/180 isn't needed since some of the code around that is different between 0.3 and 0.4, and on 0.3 `present_and_fetch` isn't segfaulting like it was before that. Correct me if I'm wrong. Easy enough to change.

notgull (approved):

ids1024:
Tagged and published.

PR #193, opened 2024-01-10
Add ability to get underlying window handle

notgull:
Adds the `get_ref` and `get_mut` functions, which can be used to get
references (mutable or otherwise) to the underlying window handle.

cc https://github.com/rust-windowing/raw-window-handle/issues/158#issuecomment-1881376603

### DIFF
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -68,13 +68,13 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> CGImpl<D, W> {
 
     /// Get the inner window handle.
     #[inline]
-    pub fn get_ref(&self) -> &W {
+    pub fn window(&self) -> &W {
         &self.window_handle
     }
 
     /// Get a mutable reference to the inner window handle.
     #[inline]
-    pub fn get_mut(&mut self) -> &mut W {
+    pub fn window_mut(&mut self) -> &mut W {
         &mut self.window_handle
     }
 
--- a/src/cg.rs
+++ b/src/cg.rs
@@ -72,12 +72,6 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> CGImpl<D, W> {
         &self.window_handle
     }
 
-    /// Get a mutable reference to the inner window handle.
-    #[inline]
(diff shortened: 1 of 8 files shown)


ids1024:
For https://github.com/iced-rs/iced/pull/2191, I also ended up needing `Arc<Window>` to create a `softbuffer::Context` and `softbuffer::Surface` from the same `HasWindowHandle + HasDisplayHandle`. Not sure how best to avoid that there.

Using `Arc` isn't really a problem even, it just seems a little odd from an API standpoint.

Having accessors like this seems reasonable, regardless.

notgull:
> For https://github.com/iced-rs/iced/pull/2191, I also ended up needing `Arc<Window>` to create a `softbuffer::Context` and `softbuffer::Surface` from the same `HasWindowHandle + HasDisplayHandle`. Not sure how best to avoid that there.

Probably by using an owned display handle, tracked in https://github.com/rust-windowing/winit/issues/3365

daxpedda (approved):
How about calling them `window()` and `window_mut()`?
See the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter).

kpreid:
In a previous much-different version of `softbuffer`, [the `window_mut()` operation that existed then had to be unsafe](https://github.com/rust-windowing/softbuffer/pull/5), because mutable access is sufficient to swap it with a completely unrelated handle, thus violating the invariant that the context referred to the expected window, allowing the window in use to be dropped. Is that not still a problem here? I have not kept up with how `raw-window-handle` works these days.

If it's still a concern, then `window_mut()` needs to be unsafe.

notgull:
Good catch! Yes that still happens, let me fix it

daxpedda (approved):

ids1024 (approved):

ids1024:
Perhaps it would also make sense to have a `.context()` accessor on `Surface`, and a `.display()` accessor on `Context`. But that is probably less often useful.

PR #196, opened 2024-01-27
Use traits to define common backend interface

ids1024:
This still uses enums to dispatch to the backends, but has the `*Dispatch` types implement the same traits. Nothing about the behavior or performance should change.

If we require backends to implement the same methods, it seems good to use traits for that. Any documentation on the interface implemented by backends can be in one place. The traits can provide default implementations where appropriate.

Hopefully this will help with more refactoring and features.


ids1024:
I see some tests are failing since an indirect dev-dependency requires Rust 1.70... not sure what we want to do there.

Though that's unrelated to this change.

notgull (changes requested):
We should move the backends into a separate directory now, since there are other non-backend files in the root of the crate.

Aside from that, LGTM.

daxpedda (approved):
LGTM!

Would love to get rid of the macro, but didn't look into it to offer any suggestions.
I second https://github.com/rust-windowing/softbuffer/pull/196#pullrequestreview-1847328480 suggestion, but I'm happy to approve it as is.

ids1024:
The alternative to the macro would be `Box<dyn ...>`, but that's probably not possible anyway with the associated type.

I think the `enum_dispatch` crate would be a popular solution here, but that adds a proc macro dependency.

notgull:
> I see some tests are failing since an indirect dev-dependency requires Rust 1.70... not sure what we want to do there.

Rebase on master for fixed MSRV CI

notgull (changes requested):
I maintain my earlier comment about the separate directory. I would move `cg.rs`, `kms.rs`, etc into a separate `sys` module to separate them from the new ones.

ids1024:
Added fixes for (unrelated) CI issues, and moved the backends to a `backends` module.

ids1024:
Added a commit that move the backend context/surface `::new` methods into traits, making it possible for `make_dispatch!` to handle `new` like it handles other methods.

notgull (approved):
Looks good to me!

PR #197, opened 2024-01-28
Set up an autorelease pipeline

notgull:
This commit adds a pipeline that automatically creates a GitHub Release
and a crates.io release when a tag is pushed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - v[0-9]+.*
+
+jobs:
+  create-release:
+    if: github.repository_owner == 'rust-windowing'
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v4
+      - uses: taiki-e/create-gh-release-action@v1
+        with:
+          changelog: CHANGELOG.md
+          branch: master
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+      - run: |
+        cargo login ${{ secrets.CRATES_IO_API_TOKEN }}
+        cargo publish
```
daxpedda, line 25: We can use [`--token`](https://doc.rust-lang.org/1.75.0/cargo/commands/cargo-publish.html#publish-options) or use `CARGO_REGISTRY_TOKEN` directly instead of `cargo login`.

daxpedda (approved):
Looks great!

PR #198, opened 2024-01-28
Fix MSRV CI issue by pinning half to v2.2.1

notgull:
This commit fixes the MSRV CI failure by pinning the half crate to
version 2.2.1. We should look for a longer term solution to this 
issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,14 +91,17 @@ jobs:
       shell: bash
       run: cargo $CMD build --verbose --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
 
+    - name: Pin versions of dev-deps
+      if: matrix.rust_version == '1.65.0'
+      run: bash tools/fix_msrv.sh
+
     - name: Build tests
       shell: bash
       if: >
         !((matrix.platform.os == 'ubuntu-latest') && contains(matrix.platform.target, 'i686')) &&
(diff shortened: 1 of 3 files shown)


daxpedda (changes requested):
I think we should pin it in CI for MSRV runs instead of in `Cargo.toml`.

Even though this is a `dev-dependency` and won't affect downstream users, it affects tests and local-development, where we are not gonna run with the same dependencies as the majority of users: just the latest.

Unfortunately this would also be more inconvenient, but I think this is necessary until https://github.com/rust-lang/cargo/issues/9930 is resolved (assuming that `half` uses correct MSRV entries in its versions).

.github/workflows/ci.yml
```diff
@@ -91,14 +91,17 @@ jobs:
       shell: bash
       run: cargo $CMD build --verbose --target ${{ matrix.platform.target }} $OPTIONS --features $FEATURES
 
+    - name: Pin versions of dev-deps
+      if: matrix.rust_version == '1.65.0'
+      run: bash tools/fix_msrv.sh
```
daxpedda, line 96: Can't we just call `cargo update -p half --precise 2.2.1`?

I would *prefer* not to use a bash script because the lack of cross-platform support kinda defeats it's purpose compared to just putting it directly into the CI script.

notgull: I didn't know about this command, thanks!

daxpedda (approved):

PR #200, opened 2024-02-10
On Orbital, fix window resize.

jackpot51:
Previously it tested if both the width AND height were changed before updating the size when it should update if width OR height were changed.


notgull (approved):
Windows error looks unrelated. Maybe it's the new Clippy dead code detector?

PR #202, opened 2024-03-04
Fix Web CI

daxpedda:
This pins `bumpalo` to fix the MSRV check in our CI.

See https://github.com/rust-windowing/raw-window-handle/pull/162, which encountered similar issues.


notgull (approved):
Thanks!

PR #210, opened 2024-04-23
Use `objc2`

### LINKED ISSUE #181: Migrate from objc to objc2

madsmtm:
Fixes #181.

This also prevents accessing the `NSView`/`NSWindow` from other threads than the main thread.

### DIFF
--- a/src/backends/cg.rs
+++ b/src/backends/cg.rs
@@ -54,8 +54,11 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for CGImpl<
             _ => return Err(InitError::Unsupported(window_src)),
         };
 
-        let mtm =
-            MainThreadMarker::new().expect("can only access AppKit handles on the main thread");
+        // `NSView` can only be accessed from the main thread.
+        let mtm = MainThreadMarker::new().ok_or(SoftBufferError::PlatformError(
+            Some("can only access AppKit / macOS handles from the main thread".to_string()),
+            None,
+        ))?;
         let view = handle.ns_view.as_ptr();
         // SAFETY: The pointer came from `WindowHandle`, which ensures that
         // the `AppKitWindowHandle` contains a valid pointer to an `NSView`.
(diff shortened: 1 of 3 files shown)


src/backends/cg.rs
```diff
@@ -45,24 +53,31 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for CGImpl<
             RawWindowHandle::AppKit(handle) => handle,
             _ => return Err(InitError::Unsupported(window_src)),
         };
-        let view = handle.ns_view.as_ptr() as id;
-        let window: id = unsafe { msg_send![view, window] };
-        let window: id = unsafe { msg_send![window, retain] };
+
+        let mtm =
+            MainThreadMarker::new().expect("can only access AppKit handles on the main thread");
```
madsmtm, line 58: Not sure if you want to handle this as a panic, as a `raw_window_handle::HandleError::Unavailable`, or something else?

src/backends/cg.rs
```diff
@@ -45,24 +53,31 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for CGImpl<
...
+                    | NSAutoresizingMaskOptions::NSViewHeightSizable.0,
+            ))
+        };
+
+        unsafe { view.addSubview(&subview) };
         let color_space = CGColorSpace::create_device_rgb();
         Ok(Self {
             layer,
-            window,
+            window: view.window().expect("view to be inside a window"),
```
madsmtm, line 80: Not sure if you want to handle this as a panic or as something else?

src/backends/cg.rs
notgull: We should add a new type for the `Error` enum to the tune of "not on the main thread."

...actually, is `Error` `non_exhaustive`? It might be better to just return a `BackendError`.

src/backends/cg.rs
notgull: I would return a `BackendError` here as well

ids1024 (approved):

notgull (approved):
Looks good to me!

PR #214, opened 2024-04-30
Bump winit to v0.30

notgull:
This commit bumps winit, which is used in some examples, to version
v0.30.

This new version comes with many new changes; mainly the move away from
closures towards applications. In order to work with this, I create a
simple "winit_app" module that allows the examples to simply and easily
create winit applications. It takes two closures: one that initializes
some state on resume, and another that handles events.

I've found this approach reduces diff noise by a lot.

Benchmarks still use deprecated functions for now.

WIP for now until I test it


Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/rust-windowing/softbuffer"
 keywords = ["framebuffer", "windowing"]
 categories = ["game-development", "graphics", "gui", "multimedia", "rendering"]
 exclude = ["examples"]
-rust-version = "1.65.0"
```
daxpedda, line 13: We should probably note that in the changelog.

Cargo.toml
```diff
@@ -79,8 +79,7 @@ cfg_aliases = "0.2.0"
 colorous = "1.0.12"
 criterion = { version = "0.4.0", default-features = false, features = ["cargo_bench_support"] }
 instant = "0.1.12"
```
daxpedda, line 81: We should probably switch to [`web-time`](https://crates.io/crates/web-time).

examples/animation.rs
```diff
@@ -7,83 +7,94 @@ use std::rc::Rc;
...
-    let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+            web_sys::window()
+                .unwrap()
+                .document()
+                .unwrap()
+                .body()
+                .unwrap()
+                .append_child(&window.canvas().unwrap())
+                .unwrap();
+        }
```
daxpedda, line 36: This can be replaced with [`WindowAttributes::with_append()`](https://docs.rs/winit/0.30.0/winit/window/struct.WindowAttributes.html#method.with_append) now.

examples/fruit.rs
```diff
@@ -4,83 +4,92 @@ use std::rc::Rc;
...
-    let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+            web_sys::window()
+                .unwrap()
+                .document()
+                .unwrap()
+                .body()
+                .unwrap()
+                .append_child(&window.canvas().unwrap())
+                .unwrap();
+        }
```
daxpedda, line 39: As above.

examples/rectangle.rs
```diff
@@ -24,97 +26,102 @@ fn redraw(buffer: &mut [u32], width: usize, height: usize, flag: bool) {
...
-                }
+            web_sys::window()
+                .unwrap()
+                .document()
+                .unwrap()
+                .body()
+                .unwrap()
+                .append_child(&window.canvas().unwrap())
+                .unwrap();
+        }
```
daxpedda, line 49: As above.

examples/winit.rs
```diff
@@ -3,75 +3,85 @@ use std::rc::Rc;
...
-    let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+            web_sys::window()
+                .unwrap()
+                .document()
+                .unwrap()
+                .body()
+                .unwrap()
+                .append_child(&window.canvas().unwrap())
+                .unwrap();
+        }
```
daxpedda, line 31: As above.

examples/winit_wrong_sized_buffer.rs
```diff
@@ -3,78 +3,88 @@ use std::rc::Rc;
...
-    }
+            web_sys::window()
+                .unwrap()
+                .document()
+                .unwrap()
+                .body()
+                .unwrap()
+                .append_child(&window.canvas().unwrap())
+                .unwrap();
+        }
```
daxpedda, line 34: As above.

examples/utils/winit_app.rs
daxpedda, line 1: Considering we are doing the exact same window creation in every example so far, would it be crazy to include that here as well?

notgull: Sounds good to me

daxpedda (approved):

PR #217, opened 2024-05-08
Implement Send and Sync for softbuffer types

### LINKED ISSUE #205: Multi-threaded
Currently, `Context` and `Surface` are neither `Send` nor `Sync`. What is the reason for this decision? Is it possible to make them at least `Send`, so they could be made `Sync` with a `Mutex`?

notgull:
Previously, types in softbuffer were !Send and !Sync. However it would
be nice if types could be shared across threads. Therefore I've made the
following changes:

- Context<D> is Send+Sync iff D is Send+Sync
- Surface<D, W> is Send iff D is Send+Sync and W is Send
- Buffer<'x, D, W> is Send iff D if Send+Sync and W is Send

Materially, I've made the following changes across the backends:

- X11, Wayland and KMS use Arc for their displays instead of Rc.
- MacOS uses MainThreadBound to secure windowing resources. This
  restriction was already implicitly enforced anyhow.
- Windows creates a thread specifically for allocating and then deallocating device contexts. This lets us get around the thread affinity problem.
- Web just isn't `Send`.

Closes #205, still WIP

### DIFF
--- a/src/backends/cg.rs
+++ b/src/backends/cg.rs
@@ -14,7 +14,7 @@ use foreign_types::ForeignType;
 use objc2::msg_send;
 use objc2::rc::Id;
 use objc2_app_kit::{NSAutoresizingMaskOptions, NSView, NSWindow};
-use objc2_foundation::MainThreadMarker;
+use objc2_foundation::{MainThreadBound, MainThreadMarker};
 use objc2_quartz_core::{kCAGravityTopLeft, CALayer, CATransaction};
 
 use std::marker::PhantomData;
@@ -30,9 +30,9 @@ impl AsRef<[u8]> for Buffer {
 }
 
 pub struct CGImpl<D, W> {
-    layer: Id<CALayer>,
-    window: Id<NSWindow>,
-    color_space: CGColorSpace,
+    layer: MainThreadBound<Id<CALayer>>,
+    window: MainThreadBound<Id<NSWindow>>,
+    color_space: SendCGColorSpace,
     size: Option<(NonZeroU32, NonZeroU32)>,
     window_handle: W,
     _display: PhantomData<D>,
@@ -84,9 +84,9 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for CGImpl<
         unsafe { view.addSubview(&subview) };
         let color_space = CGColorSpace::create_device_rgb();
         Ok(Self {
-            layer,
-            window,
-            color_space,
+            layer: MainThreadBound::new(layer, mtm),
+            window: MainThreadBound::new(window, mtm),
+            color_space: SendCGColorSpace(color_space),
             size: None,
             _display: PhantomData,
             window_handle: window_src,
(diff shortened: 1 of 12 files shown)


notgull:
@jackpot51 Please advise on Redox failures. I'm not familiar enough with Orbital's internals to know if it is thread safe or not.

notgull:
@jackpot51 Friendly reminder. I'm not familiar enough with Orbital to know how to proceed here.

src/backends/cg.rs
```diff
@@ -176,3 +179,7 @@ impl<'a, D: HasDisplayHandle, W: HasWindowHandle> BufferInterface for BufferImpl
         self.present()
     }
 }
+
+struct SendCGColorSpace(CGColorSpace);
+// SAFETY: I'm 99% sure this is thread safe.
+unsafe impl Send for SendCGColorSpace {}
```
madsmtm, line 185:
```suggestion
// SAFETY: `CGColorSpace` is immutable, and can freely be shared between threads.
unsafe impl Send for SendCGColorSpace {}
unsafe impl Sync for SendCGColorSpace {}
```

[Reference on CoreFoundation's thread safety](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/ThreadSafetySummary/ThreadSafetySummary.html#//apple_ref/doc/uid/10000057i-CH12-SW9), which CoreGraphics inherits.

madsmtm (approved):
CG impl approved with suggestion applied

jackpot51 (approved):
Orbital change looks fine

src/backends/wayland/mod.rs
```diff
@@ -136,7 +136,12 @@ impl<D: HasDisplayHandle + ?Sized, W: HasWindowHandle> WaylandImpl<D, W> {
             self.surface().commit();
         }
 
-        let _ = self.display.event_queue.borrow_mut().flush();
+        let _ = self
```
ids1024, line 139: Maybe instead of repeating this pattern in a few different places, `DisplayImpl` could have a `lock_event_queue` method?

Otherwise the Wayland changes look good.

src/lib.rs
```diff
@@ -315,3 +318,29 @@ fn display_handle_type_name(handle: &RawDisplayHandle) -> &'static str {
         _ => "Unknown Name", //don't completely fail to compile if there is a new raw window handle type that's added at some point
     }
 }
+
+#[cfg(all(not(target_arch = "wasm32"), not(target_arch = "wasm64")))]
```
daxpedda, line 322:
```suggestion
#[cfg(not(target_family = "wasm"))]
```

daxpedda (approved):
Confirming that Wasm can't be `Send` or `Sync`.

src/backends/wayland/mod.rs
notgull: I disagree. This pattern is only repeated three times, is around five lines and I'm worried that a `lock_event_queue` method would obscure the point of the code more than it would help.

notgull:
It seems that there are still bugs in the Windows backend. Sometimes sending a message to a window on another thread doesn't work. Standby while I investigate.

notgull:
> It seems that there are still bugs in the Windows backend. Sometimes sending a message to a window on another thread doesn't work. Standby while I investigate.

Okay, I've set up a new strategy for multithreading on Windows. It's a little hacky but it seems to work.

notgull:
I think we have approval from all platforms, so I'll go ahead and merge this. But I'll wait a week before releasing, that way the people using the `master` branch directly over Git have a breathing period to find bugs.

PR #223, opened 2024-06-21
Make wayland-sys dependency optional

morr0ne:
As the title says, this pr makes the wayland-sys dependency optional and only enables when the wayland feature is enabled.
I noticed that I was pulling in the wayland-sys dependency while only using only the drm backend. 
I am not sure whether or not this was intended behavior but I would assume there is no reason to pull a wayland dependency when not using wayland

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ harness = false
 [features]
 default = ["kms", "x11", "x11-dlopen", "wayland", "wayland-dlopen"]
 kms = ["bytemuck", "drm", "rustix"]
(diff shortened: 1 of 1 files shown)


notgull:
CI failures are probably not this PR's fault.

morr0ne:
> CI failures are probably not this PR's fault.

I figured as such. There seem to be a problem with the msrv and some doctests are failing because of this lint: https://rust-lang.github.io/rust-clippy/master/index.html#/needless_doctest_main

Nothing seems actually related to the pr

notgull:
Please rebase on master for the CI fixes

notgull (approved):
Thanks!

PR #226, opened 2024-06-22
Support visuals with depth 32 on X11

notgull:
This can lead to weird results in some cases, but it's better than crashing.
Once we have a better format API we can work around this.

ref rust-windowing/winit#3646


daxpedda (approved):
I assume we don't want to return an error here instead because until we add a pixel format API its not actionable for users?

notgull:
That is correct.

PR #227, opened 2024-07-10
examples: Make `utils/winit_app.rs` a `path`-referenced module again to fix autoformatting

MarijnS95:
Via `include!()` `rustfmt` doesn't know that this file belongs to the current project and doesn't autoformat it. Furthermore, jump-to-symbol likely doesn't play that well with `include!()`.

Without creating a `lib.rs` file in `utils/`, use `#[path]` to reference the existing file without following a `mod`ule hierarchy.

Note that the file remains in the `utils/` folder as that doesn't get picked up by `autoexamples`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ use winit::event::{Event, WindowEvent};
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::Window;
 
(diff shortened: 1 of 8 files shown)


notgull (approved):
Thanks!

PR #229, opened 2024-07-23
Add a MSRV policy

daxpedda:
This MSRV policy is copied from Winit.
Orbital is excluded from it because it requires nightly to begin with.

As discussed in https://github.com/rust-windowing/softbuffer/pull/228#discussion_r1686692401.


notgull (approved):
ACK

While I dislike having weird exceptions in the MSRV for certain platforms, I see the fact that it is necessary.

PR #234, opened 2024-08-26
Add support for iOS/tvOS/watchOS/visionOS

madsmtm:
A general clean-up of the CoreGraphics implementation, instead of creating a new `NSView` and rendering into the layer there, we create a sublayer. See the code comments, and also https://github.com/gfx-rs/wgpu/pull/6107 for some of the design decisions around this.

This is also a step forwards for making `softbuffer` usable from a separate thread on Apple platforms, as `CALayer` is fully thread safe (whereas `NSView`/`UIView` aren't).

Finally, it removes our dependence on having access to `NSView`, paving the way for [a future where `raw-window-handle` may only provide `CALayer`](https://github.com/rust-windowing/raw-window-handle/issues/123#issuecomment-2308505694).

Fixes https://github.com/rust-windowing/softbuffer/issues/43.

### DIFF
--- a/src/backends/cg.rs
+++ b/src/backends/cg.rs
@@ -123,9 +123,9 @@ pub struct CGImpl<D, W> {
     observer: Retained<Observer>,
     color_space: SendCGColorSpace,
     /// The width of the underlying buffer.
-    width: u32,
+    width: usize,
     /// The height of the underlying buffer.
-    height: u32,
+    height: usize,
     window_handle: W,
     _display: PhantomData<D>,
 }
@@ -249,8 +249,8 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for CGImpl<
         // by the observer using `NSKeyValueObservingOptionInitial`).
         let size = layer.bounds().size;
         let scale_factor = layer.contentsScale();
-        let width = (size.width * scale_factor) as u32;
-        let height = (size.height * scale_factor) as u32;
+        let width = (size.width * scale_factor) as usize;
+        let height = (size.height * scale_factor) as usize;
 
         Ok(Self {
             layer: SendCALayer(layer),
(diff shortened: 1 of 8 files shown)


src/backends/cg.rs
```diff
@@ -29,65 +34,231 @@ impl AsRef<[u8]> for Buffer {
...
+    ///
+    /// Can also be retrieved from `layer.superlayer()`.
+    root_layer: SendCALayer,
+    observer: Retained<Observer>,
     color_space: SendCGColorSpace,
-    size: Option<(NonZeroU32, NonZeroU32)>,
+    /// The width of the underlying buffer.
+    width: u32,
+    /// The height of the underlying buffer.
+    height: u32,
```
daxpedda, line 128: It seems that the internal API is mostly using `usize`, it might be good to store them as `usize` as well and do the conversion once instead of everywhere else.

src/lib.rs
```diff
@@ -118,7 +118,7 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> Surface<D, W> {
     /// ## Platform Dependent Behavior
     ///
     /// - On X11, the window must be visible.
-    /// - On macOS, Redox and Wayland, this function is unimplemented.
+    /// - On Apple platforms, Redox and Wayland, this function is unimplemented.
```
daxpedda, line 121: I _feel_ the term "Apple platforms" isn't particularly clear, I've never seen or used a term like this.
In Winit we also say "MacOS and iOS".

I assume you are doing this because it covers tvOS, watchOS and visionOS as well and I honestly don't have a good alternative, just thought I would leave my 2¢ here.

daxpedda (approved):
Only reviewed on a Rust level.

Probably update the README as well.

src/lib.rs
madsmtm: Indeed, I didn't want to write all the platform names. But I guess it makes more sense to do so.

madsmtm: Or maybe I'll just write AppKit and UIKit

PR #236, opened 2024-09-01
Resize the surface at the correct time in examples

madsmtm:
The examples previously showed resizing just before rendering, which is inefficient, the user should only resize when needed. Tested on macOS.

I have not updated the `winit_multithread` example, as it was a bit more cumbersome, and I wouldn't be able to test it anyways, as it doesn't seem to work on macOS.

### DIFF
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -33,12 +33,22 @@ fn main() {
         match event {
             Event::WindowEvent {
                 window_id,
+                event: WindowEvent::Resized(size),
+            } if window_id == window.id() => {
+                if let (Some(width), Some(height)) =
+                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
+                {
+                    surface.resize(width, height).unwrap();
+                }
+            }
+            Event::WindowEvent {
+                window_id,
                 event: WindowEvent::RedrawRequested,
             } if window_id == window.id() => {
-                if let (Some(width), Some(height)) = {
-                    let size = window.inner_size();
-                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
-                } {
+                let size = window.inner_size();
+                if let (Some(width), Some(height)) =
+                    { (NonZeroU32::new(size.width), NonZeroU32::new(size.height)) }
+                {
                     let elapsed = start.elapsed().as_secs_f64() % 1.0;
 
                     if (width.get(), height.get()) != *old_size {
@@ -48,7 +58,6 @@ fn main() {
 
                     let frame = &frames[((elapsed * 60.0).round() as usize).clamp(0, 59)];
 
-                    surface.resize(width, height).unwrap();
                     let mut buffer = surface.buffer_mut().unwrap();
                     buffer.copy_from_slice(frame);
                     buffer.present().unwrap();
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -47,7 +47,7 @@ fn main() {
             } if window_id == window.id() => {
                 let size = window.inner_size();
                 if let (Some(width), Some(height)) =
-                    { (NonZeroU32::new(size.width), NonZeroU32::new(size.height)) }
+                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
                 {
                     let elapsed = start.elapsed().as_secs_f64() % 1.0;
 
--- a/examples/fruit.rs
+++ b/examples/fruit.rs
@@ -23,7 +23,8 @@ fn main() {
         let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
 
         // Intentionally only set the size of the surface once, at creation.
-        // This is needed if the window chooses to ignore the size we passed in above.
+        // This is needed if the window chooses to ignore the size we passed in above, and for the
+        // platforms softbuffer supports that don't yet extract the size from the window.
         surface
             .resize(
                 NonZeroU32::new(width).unwrap(),
--- a/examples/fruit.rs
+++ b/examples/fruit.rs
@@ -20,7 +20,16 @@ fn main() {
         });
 
         let context = softbuffer::Context::new(window.clone()).unwrap();
-        let surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+        let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+
+        // Intentionally only set the size of the surface once, at creation.
+        // This is needed if the window chooses to ignore the size we passed in above.
+        surface
+            .resize(
+                NonZeroU32::new(width).unwrap(),
+                NonZeroU32::new(height).unwrap(),
+            )
+            .unwrap();
 
         (window, surface)
     })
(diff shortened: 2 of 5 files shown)


examples/rectangle.rs
```diff
@@ -45,16 +45,24 @@ fn main() {
...
+                event: WindowEvent::Resized(size),
             } if window_id == window.id() => {
-                // Grab the window's client area dimensions
-                if let (Some(width), Some(height)) = {
-                    let size = window.inner_size();
-                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
-                } {
-                    // Resize surface if needed
+                if let (Some(width), Some(height)) =
+                    { (NonZeroU32::new(size.width), NonZeroU32::new(size.height)) }
```
MarijnS95, line 51: I guess?

```suggestion
                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
```

examples/animation.rs
```diff
@@ -31,14 +31,24 @@ fn main() {
         elwt.set_control_flow(ControlFlow::Poll);
 
         match event {
+            Event::WindowEvent {
+                window_id,
+                event: WindowEvent::Resized(size),
+            } if window_id == window.id() => {
+                if let (Some(width), Some(height)) =
+                    { (NonZeroU32::new(size.width), NonZeroU32::new(size.height)) }
```
MarijnS95, line 39:
```suggestion
                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
```

examples/animation.rs
```diff
@@ -31,14 +31,24 @@ fn main() {
...
                 window_id,
                 event: WindowEvent::RedrawRequested,
             } if window_id == window.id() => {
-                if let (Some(width), Some(height)) = {
-                    let size = window.inner_size();
-                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
-                } {
+                let size = window.inner_size();
+                if let (Some(width), Some(height)) =
+                    { (NonZeroU32::new(size.width), NonZeroU32::new(size.height)) }
```
MarijnS95, line 50:
```suggestion
                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
```

examples/rectangle.rs
```diff
@@ -45,16 +45,24 @@ fn main() {
...
                     surface.resize(width, height).unwrap();
-
+                }
+            }
+            Event::WindowEvent {
+                window_id,
+                event: WindowEvent::RedrawRequested,
+            } if window_id == window.id() => {
+                // Grab the window's client area dimensions, and ensure they're valid
+                let size = window.inner_size();
```
MarijnS95, line 62: Seems a bit funky that we don't have a size query on `Surface`, is that intentional?

MarijnS95 (commented):
Does `winit` assume (for every backend) / guarantee that a `surface` is "always" created with "some" initial size, and always delivers a `Resized` event afterwards if it changed? Or otherwise, if it doesn't know/understand/have an initial size, always raises a `Resized` event?

examples/animation.rs
```diff
@@ -31,14 +31,24 @@ fn main() {
...
 
         match event {
+            Event::WindowEvent {
+                window_id,
+                event: WindowEvent::Resized(size),
+            } if window_id == window.id() => {
+                if let (Some(width), Some(height)) =
+                    { (NonZeroU32::new(size.width), NonZeroU32::new(size.height)) }
+                {
+                    surface.resize(width, height).unwrap();
```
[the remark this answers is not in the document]
madsmtm: Yeah, we probably should, but perhaps we should do that within Winit?

Actually, why are we using `NonZeroU32` again? Why can't the buffer's dimensions be zero? Do some backends panic? What if the surface is a subsurface of another surface, can it be zero then?

> minimizing

I wouldn't expect a `SurfaceResized` event on minimizing, rather something like `Occluded(true)`.

madsmtm:
> Does `winit` assume (for every backend) / guarantee that a `surface` is "always" created with "some" initial size, and always delivers a `Resized` event afterwards if it changed? Or otherwise, if it doesn't know/understand/have an initial size, always raises a `Resized` event?

Hmm, isn't this more of a Softbuffer issue? We're creating the surface _from_ the window, which itself has an initial size (the examples use `with_inner_size`).

examples/fruit.rs
```diff
@@ -20,7 +20,18 @@ fn main() {
...
+        // Intentionally only set the size of the surface once, at creation.
+        //
+        // This is not strictly needed, the size of the surface is initialized to that of the
+        // passed-in window above.
+        surface
+            .resize(
+                NonZeroU32::new(width).unwrap(),
+                NonZeroU32::new(height).unwrap(),
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
madsmtm: > It is "strictly needed" if the window doesn't conform to the requested size

Right, the window can choose to pick a different size that the requested one.

examples/rectangle.rs
madsmtm: Have opened https://github.com/rust-windowing/softbuffer/issues/237 for that.

examples/animation.rs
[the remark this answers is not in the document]
MarijnS95: > I wouldn't expect a `SurfaceResized` event on minimizing, rather something like `Occluded(true)`.

On `winit 0.29` we still see this on Windows (and maybe on X11/Wayland too, though I've only used a tiling WM/compositor without minimalization).

> Actually, why are we using `NonZeroU32` again? Why can't the buffer's dimensions be zero? Do some backends panic? What if the surface is a subsurface of another surface, can it be zero then?

I'd assume to skip rendering entirely? Not sure if it's safe/valid to call that on `.resize()`.

examples/animation.rs
MarijnS95: Looks like most were solved except this one.

examples/animation.rs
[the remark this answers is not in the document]
madsmtm: > we still see this on Windows

Yeah, seems to be a long-standing bug, see https://github.com/rust-windowing/winit/issues/783 and https://github.com/rust-windowing/winit/issues/2015.

> I'd assume to skip rendering entirely?

Hmm, yeah, though Softbuffer _could_ also just handle that internally on the platforms where rendering to a zero-sized buffer would fail, no?

madsmtm:
> > Hmm, isn't this more of a Softbuffer issue? We're creating the surface _from_ the window, which itself has an initial size (the examples use `with_inner_size`).
> 
> It seems a bit surprising that creating a `Softbuffer` surface doesn't require an initial size, as if it's able to query/follow that from the underlying window handle - but then we have to forward the size explicitly whenever a resize event occurs.

Well, on macOS/iOS at least I think we could fairly easily make the surface follow the size of the window automatically, but I don't know about the other platforms.

MarijnS95:
> Well, on macOS/iOS at least I think we could fairly easily make the surface follow the size of the window automatically, but I don't know about the other platforms.

Same on Android, by requesting a buffer size of `0` (or never setting one). I'll move the rest of the planned reply to https://github.com/rust-windowing/softbuffer/issues/237 and post it in a second!

examples/animation.rs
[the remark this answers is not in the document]
madsmtm: Discussion moved to https://github.com/rust-windowing/softbuffer/issues/238

Issue #237, opened 2024-09-01
Allow querying configured size


madsmtm:
`Surface::height` / `Surface::width` would be nice to have, instead of having to query them from the window.

MarijnS95:
We should intrinsically take #177 into account when tackling this issue. On most platforms it appears window size (appearance on screen) != surface size (physical dimensions of a pixel buffer) is natively supported (by relying on scaling hardware via the compositor) and should be exposed to the user one way or the other.

MarijnS95:
On Android we "have to" call [`ANativeWindow_setBuffersGeometry()`](https://developer.android.com/ndk/reference/group/a-native-window#anativewindow_setbuffersgeometry) to change the buffer format, "and set a size". But it explicitly documents:

> For all of these parameters, if 0 is supplied then the window's base value will come back in force.

And if I set the `width` and `height` to `0`, the `NativeWindowBufferLockGuard` / `ANativeWindow_Buffer` struct returned by `self.native_window.lock()` contains the actual size of the buffer at that point in time, and follows resizes caused by the system.
(If we set a size ourselves, #177 with hardware scaling comes into effect).

So realistically (for #237) we should have the _actual size_ available on the `Buffer` and allow `Surface` to either resize to a user-specified value, _or_ to a "special value" where it automatically follows the "window size" all the time (which the user reads back on `Buffer` whenever it's "locked for writing")?

MarijnS95:
So it looks like we have 3 situations:

1. User doesn't set (or discards/resets) a `Surface` size. They read back the size of `Buffer` every time they draw, and expect this to follow the `Window` size;
2. User specifically requests a pixel size on `Surface` (which may or may not affect the size of the window on-screen) and can expect the returned `Buffer` to match;
3. If the `Window` size doesn't change (or we have an API to control that separately, or they control it via e.g. `winit`), they get implicit (hardware) scaling?

madsmtm:
Yeah, that sounds right to me.

Question now is, do we _want_ to support all of them? Would we be fine with not having hardware scaling, and just always have the buffer be the size of the surface? So e.g. remove `resize`, and add `Buffer::width` and `Buffer::height` (in case the buffer is used concurrently with the surface being changed)?

MarijnS95:
[...]

PR #247, opened 2024-12-06
Fix examples by resizing the surface

jmjoy:
When running some examples on my Linux system, I encountered an error:

```
index out of bounds: the len is 0 but the index is 0
```

This error occurred when attempting to perform array indexing write operations on the buffer generated by `surface.buffer_mut()`.

I successfully resolved these issues by resizing the surface during initialization.

### DIFF
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -18,7 +18,14 @@ fn main() {
         let window = winit_app::make_window(event_loop, |w| w);
 
         let context = softbuffer::Context::new(window.clone()).unwrap();
-        let surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+        let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+
+        surface
+            .resize(
+                NonZeroU32::new(window.inner_size().width).unwrap(),
+                NonZeroU32::new(window.inner_size().height).unwrap(),
+            )
+            .unwrap();
 
         let old_size = (0, 0);
         let frames = pre_render_frames(0, 0);
--- a/examples/rectangle.rs
+++ b/examples/rectangle.rs
@@ -31,7 +31,14 @@ fn main() {
         });
 
         let context = softbuffer::Context::new(window.clone()).unwrap();
-        let surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+        let mut surface = softbuffer::Surface::new(&context, window.clone()).unwrap();
+
+        surface
+            .resize(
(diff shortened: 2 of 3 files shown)


MarijnS95 (changes requested):
This directly conflicts with the idea introduced in #236 (CC @madsmtm) that creating a `Surface` either infers ~a size from the given `window` _and/or_~ a `Resized` event is always delivered somewhere in application startup before the first `RedrawRequested` event.

Hacking that original misplaced `.resize()` back in is unfortunately not going to fly. We need to instead debug what your Linux sample is doing, was this on Wayland or X11?

MarijnS95:
More context is available in https://github.com/rust-windowing/softbuffer/issues/237, too.

MarijnS95:
Analyzing the code, it seems that the Wayland backend would have crashed if `resize()` is never called, because `size` contains `None`. On X11 however, without SHM support (i.e. falling back to sending all pixels over the wire), a default implementation is created with an empty `Vec`:

https://github.com/rust-windowing/softbuffer/blob/fa3a918db59c343a9a61afe09c72f0d6f9cda805/src/backends/x11.rs#L290

madsmtm:
Hmm, I'm fine with this PR? It does the same thing as the `fruit.rs` example does in https://github.com/rust-windowing/softbuffer/pull/236, i.e. resizing the surface upon initialization (since that isn't done yet internally on all platforms).

madsmtm (approved):

MarijnS95:
Then we should copy-paste that ugly edge-case, specific comment and all, to every other example?

I completely missed that we inconsistently had a weird edge-case in one but not all examples.

madsmtm:
> Then we should copy-paste that ugly edge-case, specific comment and all, to every other example?

I think so, yeah

madsmtm:
(Or we should fix the underlying issue)

MarijnS95:
> (Or we should fix the underlying issue)

(Guess why I initially "requested changes" 😉)

jmjoy:
> Hacking that original misplaced `.resize()` back in is unfortunately not going to fly. We need to instead debug what your Linux sample is doing, was this on Wayland or X11?

X11

jmjoy:
> Hmm, I'm fine with this PR? It does the same thing as the `fruit.rs` example does in #236, i.e. resizing the surface upon initialization (since that isn't done yet internally on all platforms).

In fact, I modified it based on the example of `fruit.rs`.

I also think it's ugly to add such a paragraph to each example. It would be better to have one that can fix the underlying issue.

PR #248, opened 2024-12-15
Update `minifb`'s desc in `README.md`

### LINKED ISSUE #235: Outdated information regarding the `minifb` crate  in `README.md`
> For example, it occasionally segfaults on some platforms and is missing key features such as the ability to set a window icon. While it would be possible to add these features to minifb, it makes more sense to instead use the standard window handling systems.

While I cannot confirm nor deny the part about segfaults, setting a window icon is [very much possible](https://github.com/emoon/rust_minifb/pull/274) since v0.21 (2022-03-27) and thus shouldn't be listed as one of `minifb`'s downsides.

Chubercik:
- Fixes #235.
- Simple change to remain pedantically accurate, see the PR's contents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,12 +10,12 @@ libraries that are available in the Rust ecosystem.
 Alternatives
 ==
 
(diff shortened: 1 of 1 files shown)


madsmtm (approved):
Sure

Issue #253, opened 2025-01-11
X11: Crash when resizing winit example window


FreeFull:
When running softbuffer's winit example under X11, it seems it's not guaranteed that `window.inner_size();` matches the latest processed `WindowEvent::Resized`. This can lead to an index out of bounds panic in the example's redraw code.

To reproduce, in an X11 environment:
```
cargo run --example winit
```
and then resize the window rapidly. It's also reproducible with `--release`, but not as easily. I'm not actually sure if this should be considered a bug with the example, or with winit itself.

Issue #83, opened 2023-03-31
Investigate more optimal way to implement macOS backend


ids1024:
Apparently macOS (and iOS, https://github.com/rust-windowing/softbuffer/issues/43) has a framework called [`IOSurface`](https://developer.apple.com/documentation/iosurface) for exchanging framebuffers and textures between processes, which sounds similar to the idea behind dmabufs on Linux. I think we should be use `IOSurfaces` for a front and back buffer, and use `IOSurfaceGetBaseAddress` to get a pointer to write into for no-copy presentation (https://github.com/rust-windowing/softbuffer/pull/65)? Assuming it can work with the right pixel format.

Or are there issues with this, or a better way?

ids1024:
http://russbishop.net/cross-process-rendering describes how it is possible to create an `IOSurface` with a size and format, access it from CPU, and set it as the `contents` of a `CALayer`.

* How would writing to the `IOSurface` from CPU perform? It would be good to have someone with a Mac that has a discrete GPU test this.
* Synchronization: how do we make sure the surface is no longer in use by the display server when re-using it?

ids1024:
Or possibly we could just use `CGImage` is is currently used, but with a [`CGDataProvider`](https://developer.apple.com/documentation/coregraphics/cgdataprovider) that reads from memory we can mutate? Presumably `CGImage`/`CGDataProvider` assume the memory isn't mutated, but we could do that once the provider is released. But that isn't so simple without any guarantees about when it will be released, and since we probably can't block waiting for that either.

ids1024:
So comparing these:
* `IOSurface`
  - With unified memory, this should let us write directly into the memory the GPU, to be truly no-copy. With a discrete GPU, a DMA transfer is required to get it into GPU memory. With integrated graphics on Intel macs, I think memory wouldn't be "unified" and it may need to copy from the portion of the memory allocated to the CPU to the portion allocated to the GPU?
  - Not sure how the synchronize, and make sure the `IOSurface` is no longer in use by the display server.
* `CGImage` with custom `CGDataProvider`
  - Saves the copy currently happening in the softbuffer backend, but CoreGraphics still needs to upload the data to GPU? (Into an `IoSurface` that is sent to the display server?)
     * Is there any possibility this upload could perform better than CPU access to the `IOSurface`? Presumably this is worse with unified memory, but maybe not otherwise?
  - Clear behavior with a release callback when it is no longer used by CoreGraphics
[...]

PR #254, opened 2025-01-23
Update `objc2` to `v0.6`, and use new `objc2-core-graphics` crate

madsmtm:
`objc2` now provides an interface to CoreGraphics, which has automatic memory management, and interfaces better with the other `objc2-*` crates. Note that this doesn't have the `CGDataProvider` helper we were using before, but I'd argue that's actually a benefit, since it forces us to implement the data provider ourselves (and makes it very visible that that's a place where there's room for improvement).

Another notable change is the introduction of `Retained::downcast`, which we use to ensure that our types are correct in the key-value observer (instead of blindly casting, and hoping that they are).

This requires an MSRV bump to Rust 1.71 to get access to `extern "C-unwind"` functions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust_version: ['1.70.0', stable, nightly]
+        rust_version: ['1.71.0', stable, nightly]
         platform:
           - { target: x86_64-pc-windows-msvc,   os: windows-latest,  }
           - { target: i686-pc-windows-msvc,     os: windows-latest,  }
@@ -51,7 +51,7 @@ jobs:
           - { target: wasm32-unknown-unknown,   os: ubuntu-latest,   }
         exclude:
           # Orbital doesn't follow MSRV
-          - rust_version: '1.70.0'
+          - rust_version: '1.71.0'
             platform: { target: x86_64-unknown-redox, os: ubuntu-latest }
         include:
           - rust_version: nightly
@@ -96,11 +96,9 @@ jobs:
       run: sudo apt-get install gcc-multilib
 
     - name: Pin deps that break MSRV
-      if: matrix.rust_version == '1.70.0'
+      if: matrix.rust_version == '1.71.0'
       run: |
-        cargo update -p ahash --precise 0.8.7
         cargo update -p bumpalo --precise 3.14.0
-        cargo update -p objc2-encode --precise 4.0.3
 
     - name: Build crate
       shell: bash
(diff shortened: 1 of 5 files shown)


ids1024:
Looks like the released `winit` only requires 1.70.0 (`master` requires 1.73), but the bump here is still well within the rust-windowing MSRV policy, so it seems perfectly reasonable.

> Note that this doesn't have, but I'd argue that's actually a benefit, since it forces us to implement the data provider ourselves

I think there's a word or two missing in this sentence after "doesn't have"? `CGImage` maybe?

.github/workflows/ci.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       run: sudo apt-get install gcc-multilib
 
     - name: Pin deps that break MSRV
-      if: matrix.rust_version == '1.70.0'
+      if: matrix.rust_version == '1.71.0'
       run: |
         cargo update -p ahash --precise 0.8.7
         cargo update -p bumpalo --precise 3.14.0
```
ids1024, line 102: Looks like neither of these should be needed now? And the first is causing CI errors.

madsmtm:
> > Note that this doesn't have, but I'd argue that's actually a benefit, since it forces us to implement the data provider ourselves
> 
> I think there's a word or two missing in this sentence after "doesn't have"? `CGImage` maybe?

Added "the `CGDataProvider` helper we were using before", thanks.

(I'd have linked to the helper, but [it's marked `#[doc(hidden)]`](https://docs.rs/core-graphics/0.24.0/src/core_graphics/data_provider.rs.html#41), which makes it a bit difficult).

ids1024:
It was a while ago, but I was wondering if we could do something custom with the `CGDataProvider` to swap between buffers instead of allocating each present (https://github.com/rust-windowing/softbuffer/issues/83#issuecomment-1499420842). So yeah, not having that helper in `objc2` isn't necessarily a bad thing.

Testing on an M1 Mac Mini, I see an `index out of bounds error` in the `winit` example trying to resize the window, but that seems to happen on `master` too... not something I see on Linux/Wayland, anyway. Perhaps the same issue as https://github.com/rust-windowing/softbuffer/issues/253.

src/backends/cg.rs
```diff
@@ -307,21 +290,47 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> BufferInterface for BufferImpl<'_,
...
-        );
-        let contents = unsafe { (image.as_ptr() as *mut AnyObject).as_ref() };
+        unsafe extern "C-unwind" fn release(
+            _info: *mut c_void,
+            data: NonNull<c_void>,
+            size: usize,
+        ) {
+            let data = data.cast::<u32>();
+            let slice = slice_from_raw_parts_mut(data.as_ptr(), size / size_of::<u32>());
+            // SAFETY: This is the same slice that we
```
ids1024, line 300: The end of this comment is missing.

ids1024:
I can't claim expertise with the CoreGraphics APIs, but most of this seems like a straightforward change from the old API, and the `into_raw`/`from_raw` part seems correct.

ids1024:
Ah right. Forgot to try that, and of course I have an old `Cargo.lock`. Seems to work fine after a `cargo update`.

PR #259, opened 2025-01-28
Appease clippy

madsmtm:
Fixes DRM example failing in CI [here](https://github.com/rust-windowing/softbuffer/actions/runs/12953355212/job/36132854217?pr=254).

Also pins `objc2-encode` for the MSRV CI check (will be redundant after https://github.com/rust-windowing/softbuffer/pull/254).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,7 @@ jobs:
       run: |
         cargo update -p ahash --precise 0.8.7
         cargo update -p bumpalo --precise 3.14.0
+        cargo update -p objc2-encode --precise 4.0.3
 
     - name: Build crate
       shell: bash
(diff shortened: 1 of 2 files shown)


MarijnS95 (approved):

.github/workflows/ci.yml
```diff
@@ -100,6 +100,7 @@ jobs:
       run: |
         cargo update -p ahash --precise 0.8.7
         cargo update -p bumpalo --precise 3.14.0
+        cargo update -p objc2-encode --precise 4.0.3
```
MarijnS95, line 103: Suggestion to see if we can run `cargo +nightly -Zminimal-versions generate-lockfile` before doing an MSRV test? I've been doing that on all my repositories and it at the same time ensures the lower bound of all my dependencies are correct/buildable :)

(Assuming there aren't too many different combinations / features / optional dependencies to test, of course)

madsmtm: For my part, feel free to try! I only did this now to fix CI.

PR #260, opened 2025-02-05
CI: Generate `minimal-versions` when build-testing MSRV

MarijnS95:
For https://github.com/rust-windowing/softbuffer/pull/259#discussion_r1932897555, CC @madsmtm. Not sure if we want to push ahead with this or first fix all our transitive dependencies... Lots of pain here.

Lots of crates bump their MSRV in simple semver releases, causing us to constantly have to hold back "transitive" dependencies when testing our own MSRV.  The opposite way of doing this is by immediately selecting the minimum compatible version per `Cargo.toml` semver selection using `-Zminimal-versions`, with the added benefit of validating our minimal dependency constraints in CI.

Unfortunately, even more crates than that don't validate their minimum version dependencies (barring minor improvements for "new" targets), resulting in many compilation failures and many transitive dependencies needing manual update.  Strangely, in for example the `image` crate a newer `num-traits 0.2.14` is being used but `num-bigint 0.4.0` fails to compile.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,29 @@ jobs:
       with:
         tool: wasm-bindgen-cli
 
+    - name: Install nightly rust to generate `-Zminimal-versions` lockfile (for MSRV)
+      if: matrix.rust_version == '1.71.0'
+      uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: nightly-2024-10-01
+
+    - name: Pin deps that break MSRV
+      if: matrix.rust_version == '1.71.0'
+      # By downgrading all our dependency versions, we ensure that our minimum
+      # version bounds are actually adequate (i.e. users can build `softbuffer`
+      # with minimal versions themselves) and opt-out of any unexpected MSRV
+      # bumps in semver-compatible releases of downstream crates.
+      run: |
+        cargo -Zminimal-versions generate-lockfile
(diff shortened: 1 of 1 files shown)


kpreid:
>  or first fix all our transitive dependencies... Lots of pain here.

Are you aware of `-Zdirect-minimal-versions`? It selects minimal versions only for your workspace’s direct dependencies, and normal maximal versions for all others. Thus while it doesn't strictly test every case you might care about, it won’t fail due to something out of your control.

.github/workflows/ci.yml
```diff
@@ -81,6 +81,29 @@ jobs:
       with:
         tool: wasm-bindgen-cli
 
+    - name: Install nightly rust to generate `-Zminimal-versions` lockfile (for MSRV)
+      if: matrix.rust_version == '1.71.0'
+      uses: dtolnay/rust-toolchain@master
```
madsmtm, line 86: Nit: Use `hecrj/setup-rust-action@v2` (or change both places to use the same).

madsmtm: Or maybe eschew a GitHub Action-thing, and just do `rustup install nightly-...` in the bash script below

madsmtm (approved):
I don't have a big opinion on this, but `-Zdirect-minimal-versions` sounds nice.

madsmtm (approved):

PR #262, opened 2025-02-28
fix transparency on macOS

jrmoulton:
https://developer.apple.com/documentation/coregraphics/cgimagealphainfo?language=objc

For supporting transparent windows on maOS this is all that is required. I've seen #241 which is adding alpha and it looks significantly more complicated. I might be missing why it is so complicated but this does work and I don't see why it should be any more complicated. I've tested it with tiny_skia. 

Using PremultipliedFirst instead of Last because that is consistent with the where the color channels are currently stored.

### DIFF
--- a/src/backends/cg.rs
+++ b/src/backends/cg.rs
@@ -322,8 +322,8 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> BufferInterface for BufferImpl<'_,
                 32,
                 self.imp.width * 4,
                 Some(&self.imp.color_space.0),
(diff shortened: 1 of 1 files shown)


madsmtm:
Pre-multiplied is probably wrong for ARGB, should just be `First`.

Also, this would be incompatible with our docs and other platforms, we tell you to fill the opacity with `0`, which means that with this change, the window would become completely transparent on macOS. While I'm all for progressive forward progress, I can't accept this PR as-is, sorry :/.

PR #267, opened 2025-06-05
Simplify usage of `Context`

madsmtm:
Simplify creation of `Context` in examples by constructing it before running the event loop with the new-ish [`EventLoop::owned_display_handle`](https://docs.rs/winit/0.30.11/winit/event_loop/struct.EventLoop.html#method.owned_display_handle).

This PR also makes `Context` `Clone` when the inner display handle is, to allow more easily passing it around.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Update to `objc2` 0.6.0.
 - Bump MSRV to Rust 1.71.
+- Make `Context` cloneable.
 
 # 0.4.6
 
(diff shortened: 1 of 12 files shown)


MarijnS95:
Maybe I wish this PR was split (or at least rebase-merged) in two pieces: it was a bit unclear to figure out that `Context` _was already correctly taking `HasDisplayHandle`_ (i.e. this didn't contain new changes to allow that) with the samples simply lagging behind (perhaps because there was not previously a way to store an owned provider of that, i.e. `EventLoop::owned_display_handle()` that you linked?), apart from simply deriving some clones :sweat_smile:

PR #268, opened 2025-06-05
Implement Debug for `Context`, `Surface` and `Buffer`

madsmtm:


MarijnS95 (approved):

src/backends/x11.rs
```diff
@@ -163,6 +166,7 @@ enum Buffer {
     Wire(Vec<u32>),
```
MarijnS95, line 166: Would this print the entire pixel buffer?

src/backends/android.rs
```diff
@@ -113,6 +114,7 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for Android
     }
 }
 
+#[derive(Debug)]
 pub struct BufferImpl<'a, D: ?Sized, W> {
     native_window_buffer: NativeWindowBufferLockGuard<'a>,
     buffer: Vec<u32>,
```
MarijnS95, line 120: Not sure if we want to spam the user with all pixel values :sweat_smile:

src/backends/x11.rs
madsmtm: Yeah, I guess so. Might make sense to make a small wrapper to avoid that.

Issue #238, opened 2024-09-04
Handle zero-sized buffers internally


madsmtm:
The `NonZeroU32` in our public API is ugly IMO. Is there a reason why we avoid resizing and rendering when the sizes are `0`? Does it crash on some backends? Could we instead make it be a no-op?

See also discussion in https://github.com/rust-windowing/softbuffer/pull/236#discussion_r1740169764.

madsmtm:
IIRC `CGImageCreate` will also fail if you provide 0 width/height. What I meant here was to special-case 0-sized buffers, and just no-op in `Buffer::present` (but still provide the empty `pixels` slice to users and everything).

The reason I think this makes a lot of sense is if you want to use `softbuffer` for [subsurfaces](https://github.com/rust-windowing/winit/issues/3506), as those are a _lot_ more likely to be zero-sized in some dimension.

I get that this might has weird interactions with surfaces that are tied to a window, not sure what we'd want to do there.

> I seem to recall some discussion about changing winit to return the size as `NonZeroU32`

Maybe https://github.com/rust-windowing/winit/issues/3433#issuecomment-1954061753?

MarijnS95:
In what case would a subsurface be _"a lot more likely to be zero-sized in some dimension"_? If it is for presenting something, or catching user-input in specific places, it has to have a surface area hence nonzero dimensions, right?

Matching the wayland `null` example above, should we find a way to return a `None` `Buffer` instead if/when this is the case?

Or when thinking about hardware scaling in #177 altogether, let the user request/create nonzero-sized buffers on their own regard (on platforms that support this) and present these into arbitrary sized surfaces?

MarijnS95:
Okay, makes sense.

---

Replying to `resizing` in the OP; did we discuss in a different thread/issue why a resize op needed to be "forwarded" from winit to softbuffer in the first place? Especially wrt hardware scaling?

(Well, realistically it should control the logical size of the surface in the layer stack / composition tree?)

MarijnS95:
It looks like that was discussed in https://github.com/rust-windowing/softbuffer/pull/236 first and subsequently "tracked" in https://github.com/rust-windowing/softbuffer/issues/237. If that's not adequate we should split out a new issue.

madsmtm:
> What I meant here was to special-case 0-sized buffers, and just no-op in `Buffer::present`

[...]

Issue #177, opened 2023-11-13
Hardware scaling (handle different logical vs. backing size)


jkarneges:
The readme says:

> Softbuffer should be used over pixels when its GPU-accelerated post-processing effects are not needed.

When I first read this, it didn't occur to me that "post-processing effects" included scaling. I was thinking it was more about things like CRT simulation. Unfortunately, scaling is often a requirement due to high resolution displays. Anyone doing full frame redraws and desiring good performance (60fps+) will almost certainly need to operate on a smaller buffer that can be scaled up, which `pixels` can do. I couldn't get acceptable performance with softbuffer and software scaling, although admittedly I didn't try very hard on the scaling algorithm.

I wonder if softbuffer could make an exception and support hardware scaling, if it wouldn't go against the spirit of the project. It feels like an almost table-stakes feature for most real world use.

kpreid:
Even without explicitly using GPU APIs, platforms likely have a built-in scaling operation available in their 2D graphics APIs, which could be faster due to knowing platform details, being fused with other necessary copies along the buffer-to-screen path, or actually being GPU-based.

notgull:
X11 and Windows do not support intrinsic scaling

ids1024:
On Windows we're currently using `BitBlt` (comments in SDL said this was the fastest solution when they tried it; but that was a while ago). Looks like https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-stretchblt may work here?

On macOS, would https://developer.apple.com/documentation/quartzcore/calayer/1410746-contentsscale be applicable? I'm not sure.

Anyway, if it works on some platforms, but isn't supported on others, or only sometimes is available, either:
* The API has to fail when it's not available, and users of softbuffer will need to deal with that
* `softbuffer` could have it's own code to scale in software, as a fallback.

For HiDPI uses, some platforms have a specific way to communicate what DPI scale an application should use, and for the application to indicate if it handles the scale itself or needs upscaling (for legacy software, mainly). That doesn't necessarily require handling from Softbuffer, but probably should be discussed in any documentation for this feature.

Fuseteam:
> It's not necessarily supported on all compositors. Though it probably should be. Mir doesn't seem to support it yet, anyway ([canonical/mir#2599](https://github.com/canonical/mir/issues/2599)).

It does now :D

madsmtm:
[...]

PR #269, opened 2025-06-05
Add `Buffer::width()` and `Buffer::height()` getters

madsmtm:
Part of https://github.com/rust-windowing/softbuffer/issues/237.

I've decided to return these in pixels corresponding to the surface/buffer itself, as that's what Softbuffer concerns itself with. If the user wants to render at the current screen DPI, they either need to multiply by a scaling factor themselves, or size the buffer using logical sizes and rely on the compositor to upscale it (which we don't yet expose knobs to control, but which we could in the future).

### DIFF
--- a/examples/rectangle.rs
+++ b/examples/rectangle.rs
@@ -19,7 +19,7 @@ fn redraw(buffer: &mut Buffer<'_, impl HasDisplayHandle, impl HasWindowHandle>, 
                 let red = (x & 0xff) ^ (y & 0xff);
                 let green = (x & 0x7f) ^ (y & 0x7f);
                 let blue = (x & 0x3f) ^ (y & 0x3f);
-                (blue | (green << 8) | (red << 16)) as u32
+                blue | (green << 8) | (red << 16)
             };
             buffer[(y * width + x) as usize] = value;
         }
--- a/examples/winit.rs
+++ b/examples/winit.rs
@@ -47,9 +47,9 @@ pub(crate) fn entry(event_loop: EventLoop<()>) {
                 let mut buffer = surface.buffer_mut().unwrap();
                 for y in 0..buffer.height().get() {
                     for x in 0..buffer.width().get() {
-                        let red = x as u32 % 255;
-                        let green = y as u32 % 255;
-                        let blue = (x as u32 * y as u32) % 255;
+                        let red = x % 255;
+                        let green = y % 255;
+                        let blue = (x * y) % 255;
                         let index = y * buffer.width().get() + x;
                         buffer[index as usize] = blue | (green << 8) | (red << 16);
                     }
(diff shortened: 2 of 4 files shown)


src/backends/wayland/mod.rs
```diff
@@ -263,6 +263,14 @@ pub struct BufferImpl<'a, D: ?Sized, W> {
...
 
 impl<D: HasDisplayHandle + ?Sized, W: HasWindowHandle> BufferInterface for BufferImpl<'_, D, W> {
+    fn width(&self) -> usize {
+        self.stack.member().width
+    }
+
+    fn height(&self) -> usize {
+        self.stack.member().height
+    }
+
```
madsmtm, line 273: @ids1024: I'm afraid of touching the `BorrowStack` stuff, maybe I could get you to finish the implementation here?

src/lib.rs
```diff
@@ -205,6 +205,28 @@ pub struct Buffer<'a, D, W> {
 }
 
 impl<D: HasDisplayHandle, W: HasWindowHandle> Buffer<'_, D, W> {
+    /// The amount of pixels wide the buffer is.
+    pub fn width(&self) -> usize {
```
madsmtm, line 209: I went with `usize` here, since it was just much more usable in examples.

I'd like to change `resize` to follow this direction as well, see:
https://github.com/rust-windowing/softbuffer/issues/238

src/lib.rs
```diff
@@ -205,6 +205,28 @@ pub struct Buffer<'a, D, W> {
 }
 
 impl<D: HasDisplayHandle, W: HasWindowHandle> Buffer<'_, D, W> {
```
madsmtm, line 207: I'm a bit unsure whether these should go on the buffer or the surface? Or does it make sense both places?

src/backends/wayland/mod.rs
ids1024: With the `BorrowStack` trick, we can't access both the "container" and "member", so probably it would make the most sense to just copy the `width` and `height` into fields of `BufferImpl`.

src/lib.rs
notgull: I would prefer `NonZeroU32` to accurately reflect the type information.

MarijnS95 (approved):
The examples look so much better with this change, no more "hacks" based on a possibly-stale and/or desynced window size.

Approving at least the Android parts; these getters are just fields returned internally.

src/lib.rs
MarijnS95: I'd say we should provide both independently. These are intrinsically different when (if?) we provide/support a proper abstraction for hardware scaling: #177.

src/lib.rs
madsmtm: I went with `NonZeroU32`, it's not as bad as I feared. I'll push forwards with just using `u32` in the future.

Issue #249, opened 2024-12-28
Problem with `winit_multithread` example on macOS


Veritaris:
Technical details & libs versions:
Libs:
```toml
winit = "0.30.7"
softbuffer = "0.4.6"
```
Machine: 
MacBook Pro with M3
**macOS**: `Sonoma 14.6.1`

Hi! During developing software rendering I've decided to switch making render itself in separated thread and attempted to run example from `examples/winit_multithread.rs`, but got a problem that program is "stuck" at `let size = window.inner_size();`:
```
starting thread...
making surface...
waiting for render...
```
In docs for `inner_size` is see that 
> Platform-specific
iOS: Can only be called on the main thread.

But is it also applied to `macOS` platform?
I tried to remove fetching window size and it works (I've added a little code to see that different thread rendering works):
```rust
    fn render_thread(
        window: Arc<Window>,
        do_render: mpsc::Receiver<Arc<Mutex<Surface>>>,
        done: mpsc::Sender<()>,
    ) {
        let mut c: u32 = 0u32;
        loop {
            println!("waiting for render...");
            let Ok(surface) = do_render.recv() else {
                println!("main thread destroyed");
                break;
            };

            // Perform the rendering.
            let mut surface = surface.lock().unwrap();

            let (Some(width), Some(height)) = (NonZeroU32::new(1024), NonZeroU32::new(1024)) else { todo!() };

            surface.resize(width, height).unwrap();

            let mut buffer = surface.buffer_mut().unwrap();

            for y in 0..height.get() {
                for x in 0..width.get() {
                    let red = (x + c) % 255;
                    let green = (y + c) % 255;
                    let blue = (x * y + c) % 255;
                    let index = y as usize * width.get() as usize + x as usize;
                    buffer[index] = blue | (green << 8) | (red << 16);
                }
            }

            println!("presenting...");
            buffer.present().unwrap();

            // We're done, tell the main thread to keep going.
            done.send(()).ok();
            c += 1;
        }
    }
```

madsmtm:
> > iOS: Can only be called on the main thread.
>
> But is it also applied to macOS platform?

No, it can be called from all threads on both platforms (the docs are a bit outdated on iOS), but it _will_ **block** on both platforms too, waiting on the main thread to fetch the surface size.

My recommendation is to communicate surface size changes through a channel or a mutex or something.

MarijnS95:
[...]

PR #287, opened 2025-11-23
Do not send `Window` to another thread in `winit_multithread` example

madsmtm:
The threading implications of accessing a window from another thread are unclear, so Winit chooses to block on the main thread on macOS and iOS.

This means that we have to send the surface width and height to the other thread instead of trying to access it dynamically from the window.

This fixes https://github.com/rust-windowing/softbuffer/issues/249.

### DIFF
--- a/examples/winit_multithread.rs
+++ b/examples/winit_multithread.rs
@@ -18,41 +18,34 @@ pub mod ex {
     type Surface = softbuffer::Surface<OwnedDisplayHandle, Arc<Window>>;
 
     fn render_thread(
-        window: Arc<Window>,
-        do_render: mpsc::Receiver<Arc<Mutex<Surface>>>,
+        do_render: mpsc::Receiver<(Arc<Mutex<Surface>>, NonZeroU32, NonZeroU32)>,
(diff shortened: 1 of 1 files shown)


madsmtm:
> On a related note, are such functions like `Window::inner_size()` guaranteed to return the same size until an incoming resize event has been **consumed by the event loop**?

No guarantees that I know of, no. But they should IMO, it makes no sense for you to retrieve a different window size before you've processed the resize event.

MarijnS95:
The reason I was asking is because `egui-winit` (unrelated to `softbuffer`, but interesting because we had a lot of problems with this crate thanks to that roundtrip-to-main for querying the size and other `Window` properties _on MacOS_!) queries these window properties when consuming its input `State` object via `State::take_egui_input()`: https://docs.rs/egui-winit/0.33.2/src/egui_winit/lib.rs.html#35-45

This to us means a big de-sync if the window size changes before we've handled the event (that we also pass down into `egui-winit`, no clue why they don't just collect and track those instead but there's no upstream issue yet - maybe I'll create one).

Issue #109, opened 2023-05-19
Endianness Considerations


daxpedda:
After a long discussion in https://github.com/rust-windowing/softbuffer/pull/104#discussion_r1193109522, we came to the conclusion that we should consider endianness when appropriate, this is mainly examples, tests and documentation.

Especially this line:
https://github.com/rust-windowing/softbuffer/blob/fcda747ddfb7062606787a369540db5badf03503/src/lib.rs#L340
is actually just wrong. This worked so far because literals in Rust are written in big-endian, so if you write it as 0RGB, it will come out as BGR0, because most systems are little-endian, but BGR0 is actually the correct layout.

---

I would like to mention that I'm strongly in favor of changing the current buffer from `[u32]` to `[u8]`, which seems like a much more user-friendly API to me and also makes most of the endianness considerations go away.

The main concern @ids1024 had about this is performance, which could be addressed by leaving the buffer as a `[u32]`, but expose it as `[u8]` in the `Deref` implementation, we could still introduce a `as_u32()` method then. Though I would like to see a benchmark showing an actual performance regression before we put effort into something unconfirmed like this.

Issue #207, opened 2024-03-29
Drawing on the web is very slow


GDX64:
It looks like drawing with wasm is way slower than it needs to be. I'm using tiny-skia and it takes 14ms just to present a 1000x1000 picture:

![image](https://github.com/rust-windowing/softbuffer/assets/36547504/074f6180-9e94-4fb8-8837-ab00bc08552a)

doing this I could draw way faster: 

```rust
#[cfg(target_arch = "wasm32")]
fn draw_buffer_web(win: &Window, pixmap: &Pixmap) {
    use wasm_bindgen::prelude::*;
    web_sys::console::log_1(&"draw_buffer_web".into());
    let canvas = get_a_canvas(win);
    let ctx: web_sys::CanvasRenderingContext2d = canvas
        .get_context("2d")
        .expect("Failed to get 2d context")
        .expect("Failed to get 2d context")
        .dyn_into()
        .expect("Failed to convert to CanvasRenderingContext2d");

    let width = pixmap.width();
    let clamped = wasm_bindgen::Clamped(pixmap.data());
    let image = web_sys::ImageData::new_with_u8_clamped_array(clamped, width)
        .expect("Failed to create image data");
    ctx.put_image_data(&image, 0.0, 0.0)
        .expect("Failed to put image data");

    fn get_a_canvas(win: &Window) -> web_sys::HtmlCanvasElement {
        use winit::platform::web::WindowExtWebSys;
        win.canvas().expect("Failed to get canvas")
    }
}
```

![image](https://github.com/rust-windowing/softbuffer/assets/36547504/b66921c6-1635-4768-b001-02e52ce0159d)

It looks like the current `present_with_damage`  function iterates throught the whole buffer and makes a copy each time. Wouldn't it be possible to just take in some buffer and present it with no copies or individual byte manipulation?

ids1024:
I assume this is a release build?

One problem with the web backend is that softbuffer uses BGRX color, which is converted to RGBA in the web backend. Softbuffer needs a format API to be able to use a different pixel format without conversion (https://github.com/rust-windowing/softbuffer/issues/98). Though any application that only supports rendering to one format will need conversion on some platforms, if there's not one format that's supported everywhere.

There may be other ways to improve the web backend, but its the backend I'm personally least familiar with.

GDX64:
[...]

PR #289, opened 2025-11-23
Add `Pixel` struct and `PixelFormat` enum

madsmtm:
Add:
```rust
#[repr(C)]
#[repr(align(4))]
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
pub struct Pixel {
    #[cfg(any(target_family = "wasm", target_os = "android"))]
    pub r: u8,
    #[cfg(not(any(target_family = "wasm", target_os = "android")))]
    pub b: u8,

    pub g: u8,

    #[cfg(any(target_family = "wasm", target_os = "android"))]
    pub b: u8,
    #[cfg(not(any(target_family = "wasm", target_os = "android")))]
    pub r: u8,

    pub(crate) a: u8,
}

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
pub enum PixelFormat {
    #[cfg_attr(not(any(target_family = "wasm", target_os = "android")), default)]
    Bgrx,
    #[cfg_attr(any(target_family = "wasm", target_os = "android"), default)]
    Rgbx,
}
```

This is much easier to get right than manually bit-packing into a `u32`.

Note that the format is RGBX on WASM and Android and BGRX elsewhere, which should allow moving towards zero-copying on WASM (https://github.com/rust-windowing/softbuffer/issues/207) and on Android (https://github.com/rust-windowing/softbuffer/issues/318). This is fundamentally needed because BGRA/BGRX aren't natively supported on those platforms (while RGBA/RGBX isn't guaranteed to be supported on other platforms such as Wayland and X11).

See also discussion in https://github.com/rust-windowing/softbuffer/issues/98. This is _not_ a full solution to pixel formats such as proposed by https://github.com/rust-windowing/softbuffer/pull/186 and https://github.com/rust-windowing/softbuffer/pull/241, but I believe it's an improvement we can more easily make in the meantime, and if nothing else it should at least be a stepping stone in allowing us to figure out what we _do_ want.

Fixes https://github.com/rust-windowing/softbuffer/issues/109.

Tested on:
- [x] Android
- [x] CoreGraphics
- [ ] KMS/DRM
- [ ] Orbital
- [x] Wayland
- [x] Web
- [x] Win32
- [x] X11

### DIFF
--- a/src/pixel.rs
+++ b/src/pixel.rs
@@ -1,20 +1,21 @@
-/// A RGBA pixel.
+/// A RGBA or BGRA pixel.
 ///
 /// # Representation
 ///
-/// This is a set of `u8`'s in the order BGRX (first component blue, second green, third red and
-/// last unset).
+/// This is a set of 4 `u8`'s laid out in the order defined by [`PixelFormat::default()`].
 ///
-/// If you're familiar with [the `rgb` crate](https://docs.rs/rgb/), you can treat this mostly as-if
-/// it is `rgb::Bgra<u8>`, except that this type has an alignment of `4` for performance reasons.
+/// This type has an alignment of `4` as that makes copies faster on many platforms, and makes this
+/// type have the same in-memory representation as a `u32`.
+///
+/// [`PixelFormat::default()`]: crate::PixelFormat#default
 ///
 /// # Example
 ///
 /// Construct a new pixel.
 ///
 /// ```
-/// # use softbuffer::Pixel;
-/// #
+/// use softbuffer::Pixel;
+///
 /// let red = Pixel::new_rgb(0xff, 0x80, 0);
 /// assert_eq!(red.r, 255);
 /// assert_eq!(red.g, 128);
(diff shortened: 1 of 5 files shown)


src/pixel.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    /// The alpha component.
+    ///
+    /// `0xff` here means opaque, whereas `0` means transparent.
+    ///
+    /// NOTE: Transparency is yet poorly supported, see [#17], until that is resolved, you will
+    /// probably want to set this to `0xff`.
+    ///
+    /// [#17]: https://github.com/rust-windowing/softbuffer/issues/17
+    pub a: u8,
```
madsmtm: We _could_ leave this out, but:
- There may be backends that currently read this, and leaving it as padding in the struct might cause them to read uninitialized bytes.
- It'll be useful in moving towards https://github.com/rust-windowing/softbuffer/issues/17 with less of a breaking change.

madsmtm:
> This could work with a `Pixel` implemented `bytemuck::Pod`. Though I guess `iced_tiny_skia` would still have to deal with the difference in format on the web platform.

Which is why I intentionally didn't implement such traits; I wanted there to be a bit more of a speed bump.

(Also, I didn't really want the dependency, but that's secondary).

> And if something like `iced_tiny_skia` did need to handle RGBA/BGRA differently, it would not be ideal if for iced to duplicate the `#[cfg]`s for different platforms.

Yup, that's definitely an issue.

An option for `iced_tiny_skia` (that we could explicitly recommend in our documentation) would be to do something like:
```rust
if cfg!(any(target_os = "macos", target_os = "windows", target_os = "linux")) {
    // Fast implementation on a few known platforms.
    let pixels = unsafe { std::mem::transmute::<&mut [Pixel], &mut [u32]>(pixels) };
    render(pixels);
} else {
    // Fall back to slower, copying implementation on other platforms.
    let buffer = vec![0u32; pixels.len()];
    render(&mut buffer);
    for (old, new) in pixels.zip(buffer) {
        let new = new.to_be_bytes();
        *old = Pixel::new_bgra(new[0], new[1], new[2], new[3]);
    }
}
```

Sure, it's not the most performant, but it's at least _possible_ to make performant (in contrast with the current design where `softbuffer` copies internally).

> (In particular, a non-breaking update to `softbuffer` should be able to add a new platform, or potentially change the format used on a particular backend to a more efficient one.)

Something that might help would be to provide a `PIXELS_IS_RGB_ORDER: bool` constant?

MarijnS95:
Keep in mind that any such transmuting API would have to be aware of strides (already tracked in separate issues, just mentioning it here). Fortunately on at least Android, that stride is always in the number of pixels, not in bytes (maybe for the planar API though).

madsmtm:
> Keep in mind that any such transmuting API would have to be aware of strides (already tracked in separate issues, just mentioning it here).

True, which does seem to hint that it might just be a use-case we shouldn't support.

madsmtm:
> If more than two formats are possible, it would be better to define an `enum` of different formats and have a constant/method/etc. to get what format is used. I'd probably avoid a bool.

Good point. I'll await your answer in https://github.com/rust-windowing/softbuffer/issues/98#issuecomment-3617135827, and if we decide to go with this PR, then I'll add something like:
```rust
pub enum PixelFormat {
    Rgbx,
    Bgrx,
}

/// The pixel format that `softbuffer` uses for the current platform.
pub const DEFAULT_PIXEL_FORMAT: PixelFormat = ...;
```

> As for strides: `Deref`ing to a slice of `Pixel` values won't work either if stride != width, right? So this API as is won't work if we need to support that.

Tracked in https://github.com/rust-windowing/softbuffer/issues/291. Getting `&[Pixel]` _could_ somewhat work as long as `bytestride % 4 == 0`, but downstream users such as `iced_tiny_skia` would need to be aware of the stride, yes (and if they want to expose an API to users where they aren't aware, they'd need to copy when `stride != width`).

madsmtm:
I think I'd like to go with the approach in this PR, but I still intend to later on fully fix https://github.com/rust-windowing/softbuffer/issues/98.

My idea for the end result would be to have both systems side-by-side, something like:
```rust
impl<D, W> Surface<D, W> {
    pub fn pixel_format(&self) -> PixelFormat { ... }
    pub fn configure(&mut self, width: u32, height: u32, pixel_format: PixelFormat) -> Result<(), ConfigureError> { ... }
    pub fn resize(&mut self, width: u32, height: u32) -> Result<(), ConfigureError> {
        self.configure(width, height, self.pixel_format())
    }
}

pub enum PixelFormat {
    Rgba8,
    Bgra8,
    Abgr8,
    Argb8,
    Bgra4,
    Rgb8,
    // + a bunch more
}

/// Direct buffer accessors.
impl Buffer<'_> {
    fn data(&mut self) -> &mut [u8] { ... }
    fn rows(&mut self) -> impl Iterator<Item = &mut [u8]> { ... }
}

/// Helper accessors for viewing the buffer's data as RGBA pixel data.
///
/// These are only valid when the pixel format is [`DEFAULT_PIXEL_FORMAT`], in other cases they may produce unexpected results.
///
/// See also https://github.com/rust-windowing/softbuffer/pull/313.
impl Buffer<'_> {
    fn pixels(&mut self) -> &mut [Pixel] { ... }
    fn pixel_rows(&mut self) -> impl Iterator<Item = &mut [Pixel]> { ... }
    fn pixels_iter(&mut self) -> impl Iterator<Item = (u32, u32, &mut Pixel)> { ... }
}
```

That should make the user fall into the pit of success when using `Pixel`, as it will use the most optimal format for the backend, while still supporting more complex use-cases with non-default pixel formats such as `tiny-skia` (they'd want to call `surface.configure(w, h, PixelFormat::Rgba)`, and then we'd do the copying internally for them when a backend doesn't support that pixel format).

madsmtm:
I'm going to merge this now, I belive it's the right approach for Softbuffer (it strikes a good balance between usability and performance). Even if we end up redesigning things, this is at least a step forwards toward format APIs and us being fully zero-copy.

Feel free to do "postmortem" reviews, if you spot an issue I'll gladly fix it in a follow-up.

PR #301, opened 2025-12-13
Fix documentation on docs.rs

madsmtm:
We were using the `doc_auto_cfg` feature, but that's been renamed to just `doc_cfg`.

This causes the documentation to fail building on docs.rs, see [the build logs](https://docs.rs/crate/softbuffer/latest/builds/2730258).

I'll submit a new release after this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix documentation building on `docs.rs`.
+
 # 0.4.7
 
 - Added support for Android using the `ndk` crate.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix documentation building on `docs.rs`.
```
MarijnS95: Not a user-facing change, probably not useful to have in the changelog - maybe just yank `0.4.7` given the nature of the `.8` followup?

PR #309, opened 2026-01-15
Add raytracing example

madsmtm:
Based on https://raytracing.github.io/books/RayTracingInOneWeekend.html.

I thought it'd be useful to have a more real-world example to help guide us towards which usage patterns we need to help with.


madsmtm:
I'm gonna merge this, but feel free to still comment on this PR, then I'll file a follow-up with fixes if necessary.

PR #310, opened 2026-01-17
Disable the KMS/DRM backend by default

madsmtm:
The backend was introduced in https://github.com/rust-windowing/softbuffer/pull/135 to fix https://github.com/rust-windowing/softbuffer/pull/42, but I feel like it's somewhat too niche for us to support as a "primary" platform.

In a sense, this is in preparation for some day dropping X11 as a default platform too (though that's years in the future, and should be coordinated with Winit).

Then again, maybe the cost of the extra `drm-*` dependencies isn't large enough that it really matters?

WDYT?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,9 @@ jobs:
 
         # Wayland, KMS/DRM, X11
         - target: i686-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


madsmtm:
Let's do this then, it's not a breaking change to re-add it as a default feature later if Winit adds support.

PR #311, opened 2026-01-17
Make it possible to build examples with just X11 / just Wayland

madsmtm:


MarijnS95:
> With default features enabled for Winit, it defaults to Wayland, which means it provides a `WaylandDisplayHandle`/`WaylandWindowHandle` to Softbuffer, which, when Softbuffer is compiled without support for Wayland, panics.

Yeah, exactly. That's the context I guessed from the changes, but this situation didn't immediately become clear from the title and missing description.

Indeed I meant that building `softbuffer` with a specific backend (before this fix) also had to be started/controlled with those environment variables to ensure `winit` uses a matching backend too.

Issue #291, opened 2025-11-23
Exposing stride and integration with `imgref`


madsmtm:
We currently expose a `Buffer<'_>` with `.width()`, `.height()` and `DerefMut<Target = [u32]>`, but for implementing zero-copy, we need to expose a `.stride()` method as well, see https://github.com/rust-windowing/softbuffer/pull/95. And while we're doing that, it'd be nice to add some helper methods like `.rows()` to make iteration a bit less cumbersome.

Surprise surprise, there's already a library for this! It's called [`imgref`](https://docs.rs/imgref/).

Naively, we'd like to just add `impl<'a> DerefMut<Target = Img<&'a mut [u32]> for Buffer<'a, ...>`, but that unfortunately won't work, since the way `Deref`/`DerefMut` is designed they require that we can actually produce a reference to them, and this can be somewhat difficult in this case since `Img` takes control over how the pointer is stored.

For example, consider `WaylandBuffer` as currently implemented:
```rust
struct WaylandBuffer {
    map: MmapMut,
    pub width: i32,
    pub height: i32,
    // ... Other fields.
}

impl Deref for WaylandBuffer {
    type Target = Img<&'unclear_lifetime mut [u32]>;
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut Img::new(self.map.as_mut(), self.width, self.height)
        // Won't work, the reference escapes the context.
    }
}
```

As far as I can tell, there isn't really a way to resolve this, at least not without larger changes to either Softbuffer or `imgref`?

madsmtm:
An alternative design in Softbuffer that still uses `imgref` may be something like:
```rust
impl Surface<D, W> {
    pub fn data(&mut self) -> Result<imgref::Img<Buffer<'_, D, W>>, SoftBufferError> { todo!() }
}

// Usage
let data = surface.data().unwrap();
for (y, row) in data.rows_mut().enumerate() {
    for (x, value) in row.iter_mut().enumerate() {
        let red = x % 255;
        let green = y % 255;
        let blue = (x * y) % 255;
        *value = blue | (green << 8) | (red << 16);
    }
}
data.into_buf().present().unwrap();
```

But that'd require a few changes to `imgref` to support arbitrary container types in `rows_mut`, and the `.into_buf()` feels a bit yucky.

MarijnS95:
All I know is that Android uses a pixel-stride for its directly-locked CPU-accessible `ANativeWindow`:

[...]

PR #315, opened 2026-01-19
Add `Buffer::byte_stride` and use it on Android

madsmtm:
Add:
```rust
impl Buffer<'_> {
    pub fn byte_stride(&self) -> NonZero<u32> { .. }
}
```

Fixes https://github.com/rust-windowing/softbuffer/issues/291 and allows doing full zero-copy in the future on macOS using `IOSurface` a la https://github.com/rust-windowing/softbuffer/pull/95 and on Android, see https://github.com/rust-windowing/softbuffer/issues/318.

This is also important for supporting pixel formats with a different bit depth than `32`, see https://github.com/rust-windowing/softbuffer/issues/98.

Users should prefer to use `pixel_rows` or `pixels_iter` added in https://github.com/rust-windowing/softbuffer/pull/312 to stride correctly. In case they don't, I've changed the macOS and Wayland backends to require a `stride % 16` when `cfg!(debug_assertions)`, this should make it easier to see when one is not following this requirement.

### DIFF
--- a/src/backends/cg.rs
+++ b/src/backends/cg.rs
@@ -259,8 +259,9 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for CGImpl<
     }
 
     fn buffer_mut(&mut self) -> Result<BufferImpl<'_>, SoftBufferError> {
+        let buffer_size = util::byte_stride(self.width as u32) as usize * self.height / 4;
         Ok(BufferImpl {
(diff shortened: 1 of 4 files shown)


madsmtm:
I'm gonna merge this now because I strongly believe we want stride in some shape or form, and this PR is at least a step in that direction.

If there's something I missed, feel free to comment it on this PR, then I'll fix it in a follow-up PR.

PR #320, opened 2026-01-23
Allow reading from buffers

madsmtm:
In https://github.com/rust-windowing/softbuffer/pull/313, I added a note that reading from the pixel buffer is discouraged, because I wanted to set `kIOSurfaceMapWriteCombineCache` on macOS once I got around to implementing proper `IOSurface` support. See e.g. the following links for a few places where it's used in the wild:
- WebKit on iOS: https://bugs.webkit.org/show_bug.cgi?id=130982
- Chromium (was later backed out though): https://chromium-review.googlesource.com/c/chromium/src/+/6575828

I thought this was fine, because I couldn't see a use-case where you'd need to read from the pixel buffer, but that was perhaps a bit hasty, I found at least one use-case: converting the alpha mode (premultiplying components or making the alpha channel opaque) by modifying the buffer in-place in a separate step before presenting it buffer.

This means that the argument for only having the mutable version instead of distinct `pixels/pixels_mut`, `pixel_rows/pixel_rows_mut` and `pixels_iter/pixels_iter_mut` methods is a bit weaker.

WDYT? Should we provide both immutable and mutable methods on `Buffer<'_>`?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -271,33 +271,9 @@ impl<D: HasDisplayHandle, W: HasWindowHandle> HasWindowHandle for Surface<D, W> 
 ///
 /// # Reading buffer data
 ///
-/// Reading from buffer data may perform very poorly, as the underlying storage of zero-copy
-/// buffers, where implemented, may set options optimized for CPU writes, that allows them to bypass
-/// certain caches and avoid cache pollution.
-///
-/// As such, when rendering, you should always set the pixel in its entirety:
-///
-/// ```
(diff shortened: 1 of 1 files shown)


madsmtm:
I just found https://github.com/gfx-rs/wgpu/issues/8897, we might have to make this method `unsafe` because the [write-combined memory](https://en.wikipedia.org/wiki/Write_combining) seems to work weirdly with atomics? I'm unfamiliar with how write-combined memory actually works.

PR #326, opened 2026-01-27
Add alpha conversion helpers

madsmtm:
Provide a few helper functions for handling the alpha channel when the pixel format of the buffer is `PixelFormat::Bgra8` or `PixelFormat::Rgba8`.

I'm a bit unsure whether it makes sense to provide these? Their current implementations are quite simple, and if we can provide `AlphaMode::Premultiplied` everywhere, the [un-]premultiplication functions probably aren't really necessary. And perhaps the use-case of actually needing `AlphaMode::Ignored` is kinda niche, which makes `make_pixels_opaque` unnecessary too?

---

I think the biggest problem with providing these is that it kinda signals that they're fast implementations, which... They aren't really, they could be made much more performant using SIMD and/or `rayon`.

I know that I don't want Softbuffer to be in the business of providing drawing primitives like `tiny-skia` or `vello_cpu` does, and I guess you could argue that this is kinda in that category, at least in the sense that it's a very different set of things you need to worry about (CPU speed vs. abstracting platform details and managing buffers).

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -1,0 +1,14 @@
+#[inline]
+pub(crate) fn premultiply(val: u8, alpha: u8) -> u8 {
+    // TODO: Do we need to optimize this using bit shifts or similar?
+    ((val as u16 * alpha as u16) / 0xff) as u8
+}
+
+#[inline]
+pub(crate) fn unpremultiply(val: u8, alpha: u8) -> u8 {
+    // TODO: Can we find a cleaner / more efficient way to implement this?
+    (val as u16 * u8::MAX as u16)
+        .checked_div(alpha as u16)
+        .unwrap_or(0)
+        .min(u8::MAX as u16) as u8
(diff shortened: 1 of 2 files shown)


madsmtm:
> And I'm not familiar with the tricks for optimizing this sort of thing, but I don't know if we want to add fancy SIMD code to `softbuffer` at least until `std::simd` is available to make that easy without extra dependencies...

I'm kinda in the same boat.

I _think_ we can support `AlphaMode::Premultiplied` everywhere, that seems to be the desired format on all backends, so let's wait and see until we figure that out.

madsmtm:
I'm going to go with https://github.com/rust-windowing/softbuffer/pull/361 and https://github.com/rust-windowing/softbuffer/pull/362 instead, i.e. documenting how to do conversions like this, but without explicitly providing API for it.

Issue #306, opened 2026-01-14
Is `Context` needed for anything other than performance?


madsmtm:
I've read https://github.com/rust-windowing/softbuffer/issues/37 and https://github.com/rust-windowing/softbuffer/pull/64, and from my understanding, the split between `Context` and `Surface` is purely for performance, to allow sharing resources that are expensive to create.

Is that correct?

If so, then I feel it might make sense to add a convenience wrapper like:
```rust
impl<W: HasDisplayHandle + HasWindowHandle> Surface<W, W> {
    pub fn new_single_window_bikeshed(window: W) -> Result<Self, CreationError> {
        // Conceptual implementation, needs to be done differently in reality
        // since the window needs to only be stored in the surface.
        let context = Context::new(window);
        Self::new(&context, window)
    }
}
```

And perhaps change `Surface` to be something like `struct Surface<D, W = D>` (make the two type parameters default to be the same type).

This would simplify the common use-case of having a single window to be:
```rust
struct App {
    surface: softbuffer::Surface<winit::Window>,
    // + possibly an `Option` if you want to support Android too.
}
```

While still allowing the more complex multi-window case:
```rust
struct App {
    context: softbuffer::Context<winit::OwnedDisplayHandle>,
    surfaces: Vec<softbuffer::Surface<winit::OwnedDisplayHandle, winit::Window>>,
}
```

See relatedly https://github.com/rust-windowing/softbuffer/issues/298.

Issue #222, opened 2024-06-20
using external buffer possible?


ericlangedijk:
Currently I am copying my own rgba into the buffer of the softbuffer-surface.
I was wondering if that could be faster. And if it would be possible that the softbuffer uses this external buffer as source.
I saw that in Windows the drawing is done with BitBlt (which I believe is hardware-accelerated) and also was wondering if StretchDIBits would be an option. The last one does not need a source-handle, but don't know if StretchDIBits is as optimized.

ids1024:
Could your program be refactored to write into the `softbuffer::Buffer` instead of a separate buffer?

At least on some backends, that should perform better than any possible implementation that presents from just a `&[u8]` or `&[u32]`. Not sure about the performance of `StretchDIBits`, `BitBlt`, and such. (Other than that SDL uses `BitBlt` with a comment that when implemented it ended up being faster than trying to use DirectX.)

ericlangedijk:
Yes BitBlt is fast.
I was thinking about your suggestion as well.
I could probably rewrite my home-made Bitmap (with all kinds of draw functions) to write to a 'backend' instead of its own Vec<Rgba. Lot of work but doable.
It saves a copy of some megabytes each frame....

notgull:
Many drawing interfaces have a way to render to a mutable byte slice. For instance in my own code I use [`tiny-skia::PixmapMut`](https://docs.rs/tiny-skia/latest/tiny_skia/struct.PixmapMut.html) around the softbuffer buffer in order to render in software to the  window.

ericlangedijk:
Yes I think for certain backends it will not be possible or handy. No clue about other systems.
On windows we have a u32 array which happily has exactly the same layout as my inner bitmap on which i perform all my (cpu) drawing logic buffer scaling and rendering.

daxpedda:
Maybe some platform-specific extensions would be more appropriate then.

ids1024:
If you want to write something specifically for Windows and have some very specific requirements, you can also always just do that directly with windows-rs instead of using Softbuffer.

I guess it's an open design question in softbuffer to decide when to offer platform extensions vs suggest using platform APIs directly.

notgull:
> I guess it's an open design question in softbuffer to decide when to offer platform extensions vs suggest using platform APIs directly.

I would say this is definitely a case where platform extensions should not be used. The changes required for some platforms to use direct buffers would effectively require re-architecting the entire crate.

[...]

PR #330, opened 2026-01-31
Rename lifetime `Buffer<'a>` to `Buffer<'surface>`

madsmtm:
To make it a bit clearer that the buffer borrows from the surface.

### DIFF
--- a/src/backend_dispatch.rs
+++ b/src/backend_dispatch.rs
@@ -76,7 +76,7 @@ macro_rules! make_dispatch {
 
         impl<D: HasDisplayHandle, W: HasWindowHandle> SurfaceInterface<D, W> for SurfaceDispatch<D, W> {
             type Context = ContextDispatch<D>;
-            type Buffer<'a> = BufferDispatch<'a> where Self: 'a;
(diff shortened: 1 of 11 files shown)


MarijnS95:
I thought that we wanted to detach the `Buffer` from `Surface` entirely and let the user allocate+manage+present these (not sure which concrete issue), but that fundamentally requires a separate type as it is incompatible with the current "lock()"-like APIs.

https://github.com/rust-windowing/softbuffer/issues/222#issuecomment-3568275044 appears to be somewhat related, though with the caveat that I'd still let "the `Context`" allocate a backend-specific `Buffer` that's initially unowned/dissociated from a `Surface` (https://github.com/rust-windowing/softbuffer/issues/306#issuecomment-3772246967).

madsmtm:
> I thought that we wanted to detach the `Buffer` from `Surface` entirely and let the user allocate+manage+present these (not sure which concrete issue), but that fundamentally requires a separate type as it is incompatible with the current "lock()"-like APIs.

Yeah, I'm not sure how / if we're going to square that circle, so I did this for now.

PR #363, opened 2026-04-22
Add Buffer `data_u8` and `rows_u8` accessors

### LINKED ISSUE #358: `Pixel` struct makes it hard to render directly into the surface buffer
Rendering libraries (such as `vello_cpu` or `tiny-skia`) don't know about the `Pixel` type and expect to be able to render into `&mut [u8]`. Thus the current API design with `Pixel` makes it difficult to render directly into the softbuffer buffer (although it is possible with an unsafe cast).

I think that `softbuffer` ought to encapsulate the `unsafe` and provide methods for safely accessing the surface buffer and `&mut [u8]`.

madsmtm:
Fixes https://github.com/rust-windowing/softbuffer/issues/358.


madsmtm:
After / as part of https://github.com/rust-windowing/softbuffer/pull/317, it'll probably make sense to add further methods like `data_u32`, but let's keep it simple for now.

PR #371, opened 2026-07-18
orbital: Use libredox for file syscall

willnode:
This updates orbital implementation use `libredox`, a newer stable library so that Redox can finish the syscall migration. It's already being done with winit.

Tested with running the demo on orbital: [screenshot](https://gitlab.redox-os.org/-/project/26/uploads/8151361f9c2af97366f9f1faf1f1421f/image.png) [repository](https://github.com/willnode/softbuffer/tree/libredox-demo).

Tested on:
- [X] Orbital (tier 3)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,7 +140,8 @@ web-sys = { version = "0.3.55", features = [
 
 # Orbital dependencies.
 [target.'cfg(target_os = "redox")'.dependencies]
-redox_syscall = "0.7"
+redox_syscall = "0.9"
+libredox = "0.1.18"
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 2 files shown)


bjorn3:
The syscall::fmap usage can probably be replaced by libredox::call::mmap or just libc::mmap (I think all libredox uses can go through libc instead just fine).

willnode:
Thanks, updated and tested. Managed to remove `redox_syscall` as well.

This crate uses `rustix` instead of `libc`, so `libredox` is best for now. `libc` need a lot of unsafe keyword anyway.

bjorn3:
rustix internally uses the libc crate outside Linux, so using rustix instead of libredox would work too, right?

jackpot51 (approved):

willnode:
@bjorn3 tried to implement, but relibc have no posix equivalent to `fpath` (closest one is realpath, but it's not accepting fd).

